PR #736, opened 2022-09-12
Bit field class

Sezzary:
Finished and tested.

- Implements `BitField` class object and partially adopts its usage. Full adoption can take place later.
- Adds shorter, handy typedefs for unsigned primitive types.


TombEngine/Specific/BitField.cpp
```diff
@@ -0,0 +1,286 @@
...
+			uint bit = uint(1 << i);
+			if ((packedBits & bit) == bit)
+				this->Bits[i] = true;
+		}
+	}
+
+	BitField::BitField(const string& bitString)
+	{
+		for (const char& bit : bitString)
+			this->Bits.push_back((&bit == "1") ? true : false); // TODO: Check.
```
asasas9500, line 40: Comparison should be:

`bit == '1'`

TombEngine/Specific/BitField.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+	uint BitField::GetSize()
+	{
+		return Bits.size();
+	}
+
+	uint BitField::GetCount()
+	{
+		uint count = 0;
+		for (const uint& bit : this->Bits)
```
asasas9500, line 51: this->Bits is vector of bool, so it should be:

`for (const bool& bit : this->Bits)`

asasas9500 (changes requested):

asasas9500 (dismissed):

PR #744, opened 2022-09-18
Math

Sezzary:
Just about as far as I can go with it and ready for review. I'm aware there's a lot to look through (sorry), so let's do it bit by bit. I am unsure about some decisions.

Key changes:
- Organised mathematical functions, objects, and constants into the new `Math` folder, which should perhaps be the first thing to review. It is used like a library; to access, a given file must include `Math/Math.h` in the header. Once namespace issues are sorted out, it will also require using the `TEN::Math` namespace.
- Organised container objects into their own subfolder, `Math/Containers`.
- Organised function groups with very specific purposes into their own namespaces. Namely, `TEN::Math::Random` and `TEN::Math::Geometry`.
- Replaced `GetLaraJointPosition()` and `GetAbsJointPosition()` functions with the all-purpose `GetJointPosition()`, which returns a `Vector3i` object instead of modifying an output argument.
- Renamed `PHD_3DPOS` to `Pose` as the object is now completely different from the original. While "pose" is technically the correct term and is used as a name for similar objects in other engines, as previously discussed I am very open to naming it something else. I would stress it needs to be something that specifically describes a position+orientation representation. (OpenLara calls a similar struct, used for joints, a "`Basis`".)
- Renamed `Vector2Int` to the more common `Vector2i`.
- Renamed `Vector3Int` to the more common `Vector3i`.
- Renamed `Vector3Shrt` to `EulerAngles`. This iteration is an intermediate state of the object of the same name found on the `euler_angles` branch. It adds useful constructors (from direction vector, quaternion, or rotation matrix), converters (to direction vector, quaternion, or rotation matrix), and operators.
- Napalmed `phd_global.h` and `trmath.cpp/h` out of existence.

Please check for my usual deadly typos and help me get things right. :)


rr-:
IMO each of the bullet should be a separate pull request, now it's a nightmare to review.

TombEngine/Math/Containers/BoundingBox.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "Game/animation.h"
+#include "Game/items.h"
+#include "Math/Containers/EulerAngles.h"
+#include "Math/Containers/Pose.h"
+
+//namespace TEN::Math
+//{
+	const BOUNDING_BOX BOUNDING_BOX::Zero = BOUNDING_BOX(0, 0, 0, 0, 0, 0);
+
+	BOUNDING_BOX::BOUNDING_BOX()
```
asasas9500, line 13: BOUNDING_BOX should be BoundingBox according to coding conventions.

asasas9500 (changes requested):

asasas9500 (dismissed):

PR #759, opened 2022-10-01
Input actions, GUI cleanup

Sezzary:
Still WIP! Opening PR to get started with the review process.

Input:
- Introduces the first iteration of a significant refactor of the input system, replacing legacy input bit fields with `InputAction` objects. These should serve to be simpler, more robust, and extendable.
- Adds the ability to query input "pulse" events.
- Adds the ability to query input release events.
- Global functions `IsClicked()`, `IsPulsed()`, `IsHeld()`, and `IsReleased()` may be used in place of `TrInput & IN_SOMETHING` and `DbInput & IN_SOMETHING` as transitory input polling functions until further iterations refine this new input system. Their full adoption, enabling the complete removal of `TrInput` and `DbInput`, will be done in a future PR as this will affect many files. I'm also uncertain about the names of these functions.

GUI:
- Major GUI cleanup (formatting, organisation, renamings, etc.).
- Completely removes GUI's unique method of polling inputs.
- Splits GUI object functions into their own files: GuiObjects.cpp/h.
- Ammo menu navigation loops just like in any other menu.
- PAUSE input may be used to exit the pause screen.
- Navigating long, vertical GUI lists, aggressive button taps are no longer necessary; the player may UP/DOWN to pulse the input.

Known bugs:
- Binoculars and lasersight don't work properly as I failed to correctly adopt the new input polling methods for them. I will revert my changes and use legacy input bit fields with these features for now.
- The yellow text denoting ammo count isn't drawn in inventory.


asasas9500 (dismissed):

TombEngine/Game/Lara/lara_basic.cpp
```diff
@@ -14,7 +14,7 @@
 #include "Game/Lara/lara_helpers.h"
 #include "Game/pickup/pickup.h"
 #include "Sound/sound.h"
-#include "Specific/input.h"
+#include "Specific/Input/Input.h"
```
Lwmte, line 17: As above.

TombEngine/Specific/winmain.cpp
```diff
@@ -393,7 +393,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 	Sound_Init();
 
 	// Initialise input
-	InitialiseInput(App.WindowHandle);
```
Lwmte, line 396: We use `Initialise` everywhere. Either we rename everything to `Initialize` (which I think must be done, but in other PR), or let's stick to `Initialise` for now.

TombEngine/Specific/Input/InputAction.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+		if (!this->IsHeld() || PrevTimeActive == 0.0f || TimeActive == PrevTimeActive)
+			return false;
+
+		float activeDelayInFrameTime = (TimeActive > round(initialDelayInSec / DELTA_TIME)) ? round(delayInSec / DELTA_TIME) : round(initialDelayInSec / DELTA_TIME);
+		float delayInFrameTime = std::floor(TimeActive / activeDelayInFrameTime) * activeDelayInFrameTime;
+		if (delayInFrameTime > (std::floor(PrevTimeActive / activeDelayInFrameTime) * activeDelayInFrameTime))
+			return true;
+
+		// Keeping version counting real time for future reference. -- Sezz 2022.10.01
```
Lwmte, line 61: Can you elaborate on the meaning of this commented code? Why you have left it?

TombEngine/Game/Lara/lara_basic.cpp
```diff
@@ -771,7 +772,7 @@ void lara_as_pose(ItemInfo* item, CollisionInfo* coll)
 			return;
 		}
 
-		if (TrInput & IN_WAKE)
+		if (ACTION_HELD_WAKE)
```
Lwmte, line 775: I don't fully understand the meaning of this, can you elaborate on both IN_WAKE and ACTION_HELD_WAKE meaning here?

Lwmte (commented):
I did not find any major problems with code, however, there is a weird bug I explained on Discord (if you remap look key from title flyby, camera moves a little and fadeout happens).

TombEngine/Game/Lara/lara.cpp
```diff
@@ -28,7 +28,7 @@
 #include "Game/control/volume.h"
 #include "Game/effects/lara_fx.h"
 #include "Game/effects/tomb4fx.h"
-#include "Game/gui.h"
+#include "Game/Gui.h"
```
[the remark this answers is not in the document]
Sezzary: Long ago you had talked about an intent to reformat file names this way to be more in line with TE conventions. Scripting also seems to do this, so I picked up on the practice. :)

TombEngine/Specific/winmain.cpp
Sezzary: Sure, I will make another PR for it.

TombEngine/Specific/Input/InputAction.cpp
Sezzary: The commented version relied on actual time passed (1 frame = 1 / 30), but because our delta time is a constant and can't correct for drift, occasionally a desync would occur and the pulse would stutter. To avoid this, `InputAction` counts frame time for the time being (1 frame = 1).

TombEngine/Game/Lara/lara_basic.cpp
Sezzary: I refer to a "wake" input as something that will get Lara out of the AFK pose. `IN_WAKE` was the old version for legacy input bit fields, `ACTION_HELD_WAKE` is the new one.

asasas9500 (changes requested):
Typedefs should be deleted according to conventions.

TombEngine/Game/Lara/lara.cpp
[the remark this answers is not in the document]
Lwmte: OK then

TombEngine/Specific/Input/InputAction.cpp
Lwmte: I see

TombEngine/Game/Lara/lara_basic.cpp
Lwmte: It looks like a #define, but under the hood it seems to call input action checks, I wonder if we should convert it into a class method? `if (WakeInputHappened())` or something would look more appropriate in my opinion.

Sezzary: I can do that, but probably as another loose function, as it wouldn't work as a class method. But later when inputs are tied to a particular player/controller I could do it more cleanly.

Lwmte: Yes, anyway calling it like a function seems to be better idea.

PR #765, opened 2022-10-07
Fix Monkey Crash when Monkey is triggered

goldlife2904:
If monkey is triggered, this is normally not Laras enemy.
So this condition creature->Enemy->IsLara() cant work, because creature->Enemy was previously set to nullptr which results to an uncatched exception

### DIFF
--- a/TombEngine/Objects/TR3/Entity/tr3_monkey.cpp
+++ b/TombEngine/Objects/TR3/Entity/tr3_monkey.cpp
@@ -168,11 +168,11 @@ namespace TEN::Entities::Creatures::TR3
 			AI_INFO AI;
 			CreatureAIInfo(item, &AI);
 
-			if (!creature->HurtByLara && creature->Enemy->IsLara())
+			if (!creature->HurtByLara && creature->Enemy != nullptr && creature->Enemy->IsLara())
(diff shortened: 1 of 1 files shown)


TombEngine/Objects/TR3/Entity/tr3_monkey.cpp
```diff
@@ -168,11 +168,11 @@ namespace TEN::Entities::Creatures::TR3
 			AI_INFO AI;
 			CreatureAIInfo(item, &AI);
 
-			if (!creature->HurtByLara && creature->Enemy->IsLara())
+			if (!creature->HurtByLara && creature->Enemy != nullptr && creature->Enemy->IsLara())
```
Sezzary, line 171: I would nest these conditions as a bit of security:
```
if (creature->Enemy != nullptr)
{
    if (!creature->HurtByLara && creature->Enemy->IsLara())
        creature->Enemy = nullptr;
}
```

The second change should probably stay as-is because we'd be wrongly "fixing" the logic otherwise.

goldlife2904: Had change it :-)

PR #774, opened 2022-10-13
FlyCheat configureable

goldlife2904:
To enable FlyCheat (Default):
no change in Script is needed as this is default.

To disable FlyCheat set in gameflow.lua:

Flow.EnableFlyCheat(false)


TombEngine/Scripting/Internal/ReservedScriptNames.h
```diff
@@ -109,6 +109,7 @@ static constexpr char ScriptReserved_EndLevel[]					= "EndLevel";
 static constexpr char ScriptReserved_GetSecretCount[]			= "GetSecretCount";
 static constexpr char ScriptReserved_SetSecretCount[]			= "SetSecretCount";
 static constexpr char ScriptReserved_AddSecret[]				= "AddSecret";
+static constexpr char ScriptReserved_SetFlyCheatEnabled[]		= "SetFlyCheatEnabled";
```
Sezzary, line 112: Why not simply name it `EnableFlyCheat`? :)

goldlife2904: Had change it to EnableFlyCheat option. Builders can set in gameflow.lua: Flow.EnableFlyCheat(false)

Lwmte:
I'd recommend to rename function to `ToggleFlyCheat` and also add similar function for `TogglePlayAnyLevel`

goldlife2904:
> I'd recommend to rename function to `ToggleFlyCheat` and also add similar function for `TogglePlayAnyLevel`

Yes, but it will be weird if builders has to write ToggleFlyCheat(false) - so you dont know if this "false" will Toggle "On" or Toggle "Off"... cause this are bool params it will be better to name function with words like "enable" or "disable" - it's clearly more understandable I think

PR #775, opened 2022-10-13
Rollingball with OCB 1 is silent

goldlife2904:
If Rollingball has OCB 1 the rollingball is silent. No Sounds and no Camera Shaking

### DIFF
--- a/TombEngine/Objects/TR5/Object/tr5_rollingball.cpp
+++ b/TombEngine/Objects/TR5/Object/tr5_rollingball.cpp
@@ -75,7 +75,7 @@ void RollingBallControl(short itemNumber)
 
 			if (distance < 16384)
 			{
-				if (OCB != 0x1)
+				if ((item->TriggerFlags & 1) != 1) // Flag 1 = silent.
 				{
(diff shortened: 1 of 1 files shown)


TombEngine/Objects/TR5/Object/tr5_rollingball.cpp
```diff
@@ -73,8 +75,11 @@ void RollingBallControl(short itemNumber)
 
 			if (distance < 16384)
 			{
-				Camera.bounce = -(((16384 - distance) * abs(item->Animation.Velocity.y)) / 16384);
-				SoundEffect(SFX_TR4_BOULDER_FALL, &item->Pose);
+				if (OCB != 0x1)
```
Sezzary, line 78: Since other temporary OCBs might be added, it would be best to simply check the one flag instead of the whole bit field:
```
if ((item->TriggerFlags & 1) != 1) // Flag 1 = silent.
{
    // Do stuff.
}
```

goldlife2904: Yes thats a good idea, change it :)

goldlife2904: I had change it, can you review, test and merge it?

PR #794, opened 2022-10-19
Lua Function for enable FlyBy's

goldlife2904:
Use this in Lua Script:
EnableFlyBy(ID of FlyBy)


Stranger1992:
Works nicely! All I would say now is, if you want, to try create a Node Editor entry for it. But all working fine on my end!

https://user-images.githubusercontent.com/84292688/197420444-90742b67-d4b9-48d9-9972-7ea538749b28.mp4

```lua
LevelFuncs.FlyBytest = function()
    --IF CTRL is pressed then enable FlyBy 7
        if KeyIsHit(8) then
            EnableFlyBy(7)
        end
    end
```

PR #804, opened 2022-10-25
Fix Underwaterdoor, Fix Doubledoors

goldlife2904:
Underwater and double doors now can't still be interacted even if they are opened


Kubsy:
Double doors work :)

Kubsy:
underwater door works

PR #806, opened 2022-10-26
Lua Function for Poison Lara

goldlife2904:
As there is no functionality to poison Lara, and there is (at this time) no "Lara" Lua Class, I had integrate it to Moveable Objects.

To Poison Lara write Lua Function:

**SetPoison(int $potency)**

and to Remove, simple call function:

**RemovePoison()**


TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
```diff
@@ -18,6 +18,7 @@
 #include "Logic/LevelFunc.h"
 #include "Rotation/Rotation.h"
 #include "Vec3/Vec3.h"
+#include "lara.h"
```
Sezzary, line 21: Prefer to use the full include path: "Game/Lara/lara.h". Also please keep includes ordered alphabetically, as an effort was recently started to do that across the codebase. :)

goldlife2904: done ;-)

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.h
```diff
@@ -101,6 +101,8 @@ class Moveable : public NamedBase<Moveable, short>
 	void MakeInvisible();
 	void Explode();
 	void Shatter();
+	void SetPoison(int $potency);
```
hispidence, line 104: Remove the `$` in front of the arg name, as it will probably confuse people into thinking it has some special behaviour in C++ (I didn't even think it was legal to start an identifier with a dollar sign)

hispidence (changes requested):

Issue #294, opened 2021-08-15
Coding conventions


Lwmte:
## General Rules

* **Do not silently comment out any lines or change operators or conditional statements**, even for testing purposes. Should you comment out certain code, also leave a comment beginning with `TODO` prefix, and date-author signature, like this:

``` c
    // TODO: This code caused NPCs to run into walls, so it is disabled for now. -- Lwmte, 21.08.19
    // GetBoundsAccurate(blah);
```

* **Do not modify existing code without a clear understanding of what it does.** Even if you think you are fixing bugs or restoring original behavior by reverting the code, consult the TombEngine discord before committing any changes. It is possible you may ruin the work of other people since the original Core Design codebase is very fragile and it is very easy to falsely identify code as bugged.

* **Do not introduce quick hacks to fix issues**. Hacks like `if (room == 314)` or `if (direction == WEST)` should be left in the history of bad Core Design coding practices.

* If code remains unimplemented or you need to visit another part of the code for a considerable amount of time to implement missing functionality for that first part of code, **leave a comment** prefixed `FIXME` and with a date-author signature, describing missing functionality that you are about to add:

``` c
    // FIXME: I need to precisely calculate floor height, but for now let's keep original.  -- Lwmte, 21.08.19
    foo = GetFloor(blah);
```

* Make sure that **any new code is warning-free** before committing. Exceptions to this guideline include cases where fixing warnings would require you to make many changes outside of the new code, where the warning originates from a 3rd-party library, or where fixing the warning would make the code significantly less readable or performant.

* Avoid using magic numbers. Use constants or enums instead.

* Use **American English**.

## Parenthesis and new lines

Please do not write like this:

``` c
if (blah) {
    foo;
    foo_foo;
} else {
    bar;
    bar_bar; }
```

Write like this instead:

``` c
if (blah)
{
    foo;
    foo_foo;
} 
else
{
    bar;
    bar_bar;
}
```

However, if you have only one line enclosed (i. e. only `foo`), and there are no following `else` or `else if` conditions, you may omit brackets and do it like this:

``` c
if (blah)
    foo;
```

For one-line if statements, if the condition itself is multi-line, brackets should be added anyway to maintain readability:

```
if (item.RoomNumber == 123 || IsHeld(In::Action) &&
    LaraHasEatenSandwich == SANDWICH_BACON)
{
        CurrentLevel = 4;
}
```

Same rule applies to loops.

[...]

PR #811, opened 2022-11-02
Further weapon code cleanup

Sezzary:


TombEngine/Game/Lara/lara_fire.cpp
```diff
@@ -33,23 +33,46 @@
 using namespace TEN::Entities::Generic;
 using namespace TEN::Input;
 using namespace TEN::Math;
+using std::pair;
```
Lwmte, line 36: I am generally against shortening `std` types, because seeing `std::pair` in one place and `pair` in another can confuse someone.

TombEngine/Game/Lara/lara_fire.cpp
```diff
@@ -386,9 +501,7 @@ void LaraGun(ItemInfo* laraItem)
 	}
 
 	if (laraItem->HitPoints <= 0)
-	{
```
Lwmte, line 389: Removing brackets in this case decreases readability, I think I will add a new rule to coding conventions for this particular case. Can you revert it?

TombEngine/Objects/TR3/Vehicles/upv.cpp
```diff
@@ -227,7 +227,7 @@ namespace TEN::Entities::Vehicles
 		auto UPV = GetUPVInfo(UPVItem);
 		auto* lara = GetLaraInfo(laraItem);
 
-		auto& ammo = GetAmmo(laraItem, LaraWeaponType::HarpoonGun);
+		auto& ammo = GetAmmo(*lara, LaraWeaponType::HarpoonGun);
```
Lwmte, line 230: Is this correct? I am not familiar with pointers/references in C++, but what exactly is happening here? Making Lara pointer a reference?

TombEngine/Game/Lara/lara_fire.cpp
Sezzary: Oh, I have a habit of doing this. I can avoid it from now on and gradually revert wherever it's been done already.

TombEngine/Objects/TR3/Vehicles/upv.cpp
Sezzary: I learned that it's better favour references over pointers where possible, and I've been gradually pulling conversions through. It might take a while so there will be some of this funny syntax sitting around while it happens.

TombEngine/Game/Lara/lara_fire.cpp
Sezzary: Sure, I can adapt.

TombEngine/Objects/TR3/Vehicles/upv.cpp
Lwmte: OK, I got it, thanks!

TombEngine/Game/Lara/lara_fire.cpp
Lwmte: Thanks, I have also added a point to #294 about that.

TombEngine/Game/Lara/lara_fire.cpp
Lwmte: Yes please, generally shortening `std` types/methods is considered a bad practice, if I recall.

PR #820, opened 2022-11-04
Uw door switch fix

Kubsy:


Sezzary:
PR also includes an example of a proposal to change to a more unified unit system. Instead of `CLICK()` and `SECTOR()`, we could measure everything in terms of a `BLOCK()`. For convenience, the function may take a divisor argument, where `BLOCK(1, 2)` reads as "one half of a block", and `BLOCK(1, 4)` reads as "one quarter of a block". Using the divisor argument instead of writing something like `BLOCK(0.25f)` ensures the return value is an `int`.

Sezzary (approved):

PR #824, opened 2022-11-08
Strings reformatting

Sezzary:
Menu strings from TRFan94 on Discord reformatted for consistency + extra corrections.


Lwmte:
I don't really think we should bring back empty lang strings while they are all empty. A comment may be added to strings file in the beginning, noting that it's possible to expand string entry tables with extra language variations, but that's all I'd do.

PR #825, opened 2022-11-08
"None" state for creatures

Sezzary:
Motivation:
`RequiredState` is used exclusively by creature entities to automate strings of dispatches between two distant states. In `AnimateItem()`, it was set to 0 by default, as Core had originally reserved state 0 to mean "none". This means creature state enums often skipped 0 and started at 1, or loosely labeled 0 as `XYZ_CREATURE_STATE_NONE`, adding a lot of redundancy. However, this convention wasn't enforced, and many creatures use state 0 for whatever they want anyway where either the `RequiredState` mechanism went unused or perhaps a dev didn't know how the system worked.

This PR designates the constant `NO_STATE` (equal to -1) as the default "none" state. I now understand the issue with having invalid states (they address the symptom and not the cause), but it should be safe here as it doesn't actually change anything about the existing system. `NO_STATE` is *consistently* as invalid as 0 *sometimes* was.

Additional changes:
- Remove `WALL_SIZE` constant.


TombEngine/Game/collision/collide_room.cpp
```diff
@@ -150,7 +150,7 @@ CollisionResult GetCollision(Vector3i pos, int roomNumber, short headingAngle, f
 	int adjacentRoomNumber = GetRoom(location, pos.x, point.y, pos.z).roomNumber;
 	return GetCollision(point.x, point.y, point.z, adjacentRoomNumber);
 
-	Random::TestProbability(0.5f);
+	Random::TestProbability(1.0f / 2);
```
asasas9500, line 153: Why?

TombEngine/Objects/TR2/Entity/tr2_bird_monster.cpp
```diff
@@ -7,25 +7,24 @@
...
+using namespace TEN::Math;
 
 namespace TEN::Entities::Creatures::TR2
 {
 	constexpr auto BIRD_MONSTER_ATTACK_DAMAGE = 200;
 	constexpr auto BIRD_MONSTER_SLAM_CRUSH_ATTACK_RANGE = SQUARE(SECTOR(1));
 	constexpr auto BIRD_MONSTER_PUNCH_ATTACK_RANGE		= SQUARE(SECTOR(2));
 
-	#define BIRD_MONSTER_WALK_TURN_RATE_MAX ANGLE(4.0f)
+	const auto BIRD_MONSTER_WALK_TURN_RATE_MAX = ANGLE(4.0f);
```
asasas9500, line 22: Should be `constexpr`?

TombEngine/Objects/TR5/Entity/tr5_submarine.cpp
```diff
@@ -422,13 +422,13 @@ namespace TEN::Entities::Creatures::TR5
 				pos = chaffItem->Pose.Position;
 				item->Animation.ActiveState = pos.x / 4;
 				item->Animation.TargetState = pos.y / 4;
-				item->Animation.RequiredState = pos.z / 4;
+				item->Animation.RequiredState = ((pos.z / 4) == 0) ? NO_STATE : (pos.z / 4);
```
asasas9500, line 425: Should be reverted since this `RequiredState` does not contain a state.

asasas9500 (changes requested):

TombEngine/Game/collision/collide_room.cpp
Sezzary: Ask Lwmte.

TombEngine/Objects/TR2/Entity/tr2_bird_monster.cpp
Sezzary: `ANGLE()` is not a lambda, so this line only works with `const`. Good enough for now.

asasas9500: Please apply `constexpr` to all functions in `TombEngine/Math/Legacy.cpp` and then to all `ANGLE()` calls.

TokyoSU:
this: Remove WALL_SIZE constant.
won't be accepted, it's used everywhere and now with #898 monty use it for room collector.

Sezzary:
> this: Remove WALL_SIZE constant. won't be accepted, it's used everywhere and now with #898 monty use it for room collector.

Search and replace. WALL_SIZE shouldn't exist.

PR #845, opened 2022-11-14
Lua Lara Functions for Poison

goldlife2904:
To Poison Lara write Lua Function:

`SetPoison(int potency)`

to get actual Poison Potency write

`GetPoison()`

and to Remove completely poison of Lara, simple call function:

`RemovePoison()`


TombEngine/Scripting/Internal/TEN/Objects/Lara/LaraObject.cpp
```diff
@@ -45,11 +45,48 @@ bool LaraObject::GetOnFire() const
...
+	auto* lara = GetLaraInfo(m_item);
+
+	return lara->PoisonPotency;
+}
+
+/// Remove Poison
+// @function Moveable:RemovePoison
+// @usage
+// Lara:RemovePoison()
+void LaraObject::RemovePoison()
```
Lwmte, line 75: I doubt that we need dedicated  `RemovePoison` function, because in that case similar functions should be introduced for every pair of getter/setter that is already existing in API.

PR #847, opened 2022-11-14
Lua functions SetAir() and GetAir()

goldlife2904:
Add SetAir(int air) and GetAir() to LaraObject Class.


TombEngine/Scripting/Internal/TEN/Objects/Lara/LaraObject.cpp
```diff
@@ -79,6 +80,30 @@ void LaraObject::RemovePoison()
 	lara->PoisonPotency = 0;
 }
 
+/// Set Air of Lara
+// @function LaraObject:SetAir
+// @tparam Air Value 
+// @usage
+// Lara:SetAir(100)
+// Max Value: 1800
+void LaraObject::SetAir(int air = 0)
```
Lwmte, line 89: Does `air = 0` here makes sense? I thought that for optional arguments, you should use `std::optional`, but in this case it's not applicable (why lack of parameter should zero out air anyway?).

PR #862, opened 2022-11-21
Lua set level secrets

goldlife2904:
Builders can now decide if their are level secrets displaying additionally in statistic screen.
If yes they can add this line into their level script:

MyLevel = Level.new()
MyLevel.secrets = 3


TombEngine/Scripting/Internal/TEN/Flow/Level/FlowLevel.h
```diff
@@ -38,6 +38,7 @@ struct Level : public ScriptInterfaceLevel
 	int LevelFarView{ 0 };
 	bool UnlimitedAir{ false };
 	std::vector<InventoryItem> InventoryObjects;
+	short LevelSecrets{ 0 };
```
Lwmte, line 41: I think you can make this property `int`, no need in using `short` type anywhere in new code.

TombEngine/Scripting/Internal/TEN/Flow/Level/FlowLevel.h
```diff
@@ -55,4 +56,6 @@ struct Level : public ScriptInterfaceLevel
 	short GetFogMinDistance() const override;
 	short GetFogMaxDistance() const override;
 	short GetFarView() const override;
+	void SetSecrets(short secrets);
```
Lwmte, line 59: Are these necessary, if you're directly setting secrets field from lua?

goldlife2904: no. is not necessary for lua. but perhaps it could be necessary in code, / c++ so I had leave this function.

TombEngine/Scripting/Internal/TEN/Flow/Level/FlowLevel.h
goldlife2904: done

PR #881, opened 2022-12-01
Lua additions for 1.0.4

Lwmte:
* Added `GetVehicle` and `GetTarget` script commands for `Lara` object. They return `Moveable` object (if such object - target or vehicle - exists), which can be used as any other moveable (e.g. `GetName`, `SetHP` etc).
* Mounted vehicle can now activate volumes which have Lara activator type. In such case, vehicle will be passed as an activator, not Lara.
* Add `ActionID` enum which can now be used instead of numericals in `KeyIsHeld` etc functions (e.g.: `Misc.KeyIsHeld(Misc.ActionID.FORWARD)` instead of `Misc.KeyIsHeld(0)`).
* Fix mounted vehicles ignoring `Disable`, `Shatter` and `Explode` script commands.


Kubsy:
GetVehicle(), works with SetColour, SetVelocity (whoever made this function, thank you so much), SetRotation etc but it doesnt work with things like shatter or explode (not sure if intended) 

will now try with GetTarget()

Kubsy:
GetTarget() partially works, I tested with SetHP() and Shatter() and both worked. I tried with SetVelocity() and SetColor() and it didnt work

TombEngine/Scripting/Internal/TEN/Objects/Lara/LaraObject.cpp
```diff
@@ -10,6 +10,7 @@
 #include "Game/effects/item_fx.h"
 #include "Specific/level.h"
 #include "ReservedScriptNames.h"
+#include "Scripting/Internal/TEN/Objects/ObjectsHandler.h"
```
hispidence, line 13: Is this needed or a holdover from when you were using GetByName?

Lwmte: Yes, I have removed it :)

Lwmte:
> GetVehicle(), works with SetColour, SetVelocity (whoever made this function, thank you so much), SetRotation etc but it doesnt work with things like shatter or explode (not sure if intended)
> 
> will now try with GetTarget()

I have added a workaround which allows to explode/shatter mounted vehicles. It's not beautiful, but it should also solve problems when vehicle is removed/destroyed/forcefully set to inactive state from scripts - in this case, Lara now will automatically dismounted from vehicle.

Kubsy:
Everything works so approving

Kubsy (approved):

PR #882, opened 2022-12-02
refractored charging sparks and charging fire for enemies

Nemoel-Tomo:
fixed enemies: seth and harphy

### DIFF
--- a/TombEngine/Game/animation.cpp
+++ b/TombEngine/Game/animation.cpp
@@ -502,19 +502,3 @@ Vector3i GetJointPosition(ItemInfo* item, int jointIndex, const Vector3i& offset
 	return Vector3i(pos);
 }
 
-SparkOffsets GetSparkOffset(Vector3i pos)
-{
-	SparkOffsets pos1;
-
-	short random = GetRandomControl();
-	
-	pos1.x = (GetRandomControl() & 0x7FF) + pos.x - SECTOR(1);
-	pos1.y = (GetRandomControl() & 0x7FF) + pos.y - SECTOR(1);
-	pos1.z = (random & 0x7FF) + pos.z - SECTOR(1);
-
-	pos1.vx = 8 * (pos.x - pos1.x);
-	pos1.vy = 8 * (pos.y - pos1.y);
-	pos1.vz = 8 * (SECTOR(1) - (random & 0x7FF));
-
-	return pos1;
-}
--- a/TombEngine/Game/animation.cpp
+++ b/TombEngine/Game/animation.cpp
@@ -501,3 +501,20 @@ Vector3i GetJointPosition(ItemInfo* item, int jointIndex, const Vector3i& offset
 	g_Renderer.GetItemAbsBonePosition(itemNumber, pos, jointIndex);
 	return Vector3i(pos);
 }
+
+SparkOffsets GetSparkOffset(Vector3i pos)
+{
+	SparkOffsets pos1;
+
+	short random = GetRandomControl();
+	
+	pos1.x = (GetRandomControl() & 0x7FF) + pos.x - SECTOR(1);
+	pos1.y = (GetRandomControl() & 0x7FF) + pos.y - SECTOR(1);
+	pos1.z = (random & 0x7FF) + pos.z - SECTOR(1);
+
+	pos1.vx = 8 * (pos.x - pos1.x);
+	pos1.vy = 8 * (pos.y - pos1.y);
+	pos1.vz = 8 * (SECTOR(1) - (random & 0x7FF));
+
+	return pos1;
+}
--- a/TombEngine/Game/animation.h
+++ b/TombEngine/Game/animation.h
@@ -22,6 +22,16 @@ enum class AnimCommandType
 	Flipeffect
 };
 
+struct SparkOffsets
+{
+	int x;
+	int y;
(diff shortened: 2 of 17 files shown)


TombEngine/Game/animation.h
```diff
@@ -22,6 +22,16 @@ enum class AnimCommandType
 	Flipeffect
 };
 
+struct SparkOffsets
+{
+	int x;
```
Lwmte, line 27: 1) Don't use `int` for types with three components, such as position or velocity. We have special `Vector3i` type for that.
2) `SparkOffsets`, should it ever exist (which I am not sure), should not be placed in `animation.h`.

TombEngine/Game/animation.cpp
```diff
@@ -501,3 +501,20 @@ Vector3i GetJointPosition(ItemInfo* item, int jointIndex, const Vector3i& offset
 	g_Renderer.GetItemAbsBonePosition(itemNumber, pos, jointIndex);
 	return Vector3i(pos);
 }
+
+SparkOffsets GetSparkOffset(Vector3i pos)
```
Lwmte, line 505: This function does not belong to `animation.cpp`. Find better place, probably `effects.cpp`.

TombEngine/Game/effects/effects.cpp
```diff
@@ -1705,3 +1705,42 @@ void ProcessEffects(ItemInfo* item)
 	if (item->IsLara() && GetLaraInfo(item)->Control.WaterStatus == WaterStatus::FlyCheat)
 		item->Effect.Type = EffectType::None;
 }
+
+void TriggerAttackFlame(int x, int y, int z, Vector3 color, int size)
```
Sezzary, line 1709: - Use `Vector3` container. Effects shouldn't need integer positions, this is a Core relic. Later we can refactor particle structs to take advantage of this.
- Make parameters read-only references.

`void TriggerAttackFlame(const Vector3& pos, const Vector3& color, int size)`

TombEngine/Game/effects/effects.cpp
```diff
@@ -1705,3 +1705,42 @@ void ProcessEffects(ItemInfo* item)
...
+{
+	auto* spark = GetFreeParticle();
```
Sezzary, line 1711: Adopt the use of references. Calling old functions, for now, it can be done like this: `auto& spark = *GetFreeParticle();`

TombEngine/Game/animation.h
```diff
@@ -22,6 +22,16 @@ enum class AnimCommandType
 	Flipeffect
 };
 
+struct SparkOffsets
```
Sezzary, line 25: This struct and the `GetSparkOffsets()` function that returns it aren't really necessary, there are simpler ways of doing this. We can discuss on Discord. :)

TombEngine/Game/effects/spark.cpp
```diff
@@ -140,4 +142,32 @@ namespace TEN::Effects::Spark
 			s.active = true;
 		}
 	}
+
+	void TriggerAttackSpark(SparkOffsets* pose, Vector3 color)
```
Sezzary, line 146: As with the struct, let's discuss on Discord.

Sezzary, line 146: Also colour should be a const reference: `const Vector3& color`.

TombEngine/Game/animation.h
Nemoel-Tomo: I replaced it into effects.h

TombEngine/Game/animation.cpp
Nemoel-Tomo: I replaced it into effects.cpp

TombEngine/Game/animation.h
Nemoel-Tomo: and I changed it to Vector3i

TombEngine/Game/effects/effects.cpp
Nemoel-Tomo: changed

TombEngine/Game/animation.h
Nemoel-Tomo: I needed a container with 6 arguments

TombEngine/Game/effects/spark.cpp
Nemoel-Tomo: changed

TombEngine/Game/animation.h
Sezzary: I did a little tinkering, and in fact, it seems we don't even need the struct. We can do everything in `TriggerAttackSpark()`, which is very convenient. The steps are a bit much to convey through text so I will push something in 8 hours to show you. :)

Kubsy:
https://github.com/MontyTRC89/TombEngine/pull/885 please test this before merging this branch to master

TombEngine/Objects/TR4/Entity/tr4_setha.cpp
```diff
@@ -368,7 +286,7 @@ namespace TEN::Entities::TR4
 				}
 				else
 				{
-					if (creature->ReachedGoal)
+					if (!creature->ReachedGoal)
```
Lwmte, line 289: Why this condition was inverted?

PR #904, opened 2022-12-10
Weapon handling cleanup

Sezzary:
Fourth iteration aimed toward making sense of weapon code for future benefits when rewriting it. Reduces copy-paste, simplifies some logic, and conversions from pointer to reference arguments continue to be pushed farther up the call hierarchy.


Lwmte:
Not approving because there is unneeded conversion to references.

Issue #893, opened 2022-12-07
Implement activator queue for volumes


Lwmte:
Currently it's only possible to activate volume using only a single activator. As long as there is applicable activator inside volume, its state will be always set as "busy", until activator leaves the volume. This prevents using volumes in cases when several objects are overlapping inside of it (e.g. if volume overlaps animating, but should also activate rollingball). This can be solved by implementing activator queue, which will remember state for every incoming activator, not just a single one.

PR #906, opened 2022-12-11
Volume system refactor

Lwmte:
Solves #893 and implements activator queue for volumes, which allows infinite amount of objects to activate same volume at the same time. Solves issues with multiple NPCs not reacting to same volume (e.g. @GeckoKid-JW subway escalator case).

Additionally, this PR introduces script methods for volumes and also changes intersection method of volumes with player, so rough box is used instead precise one, which prevents false "oscillating" containment events from happening.

**Things to test: **

- Check if previous volume setups work as before.
- Check if multi-activator volume setups now work properly (including branching cases, when different activators should activate different node paths or script branches).
- Check if volume states are correctly saved / reloaded in savegames.
- Check if volume script functions work as intended.


Kubsy:
works, even on save/load and loading from title.

PR #907, opened 2022-12-12
Fix tr5 autoguns

Nemoel-Tomo:


TombEngine/Objects/TR5/tr5_objects.cpp
```diff
@@ -14,26 +14,26 @@
...
-#include "tr5_imp.h"		  // OK
-#include "tr5_lagoon_witch.h" // OK
-#include "tr5_larson.h"		  // OK
-#include "tr5_laser_head.h"	  // OK
-#include "tr5_lion.h"		  // OK
-#include "tr5_reaper.h"		  // OK
-#include "tr5_roman_statue.h" // OK
-#include "tr5_submarine.h"	  // OK
-#include "tr5_willowwisp.h"	  // OK
+#include "Objects/TR5/Entity/AutoGun.h"			 // OK
```
Kubsy, line 17: Shouldn't it be TR5_AutoGun or are we considering the TR3 version as the same?

Kubsy: okay :)

PR #912, opened 2022-12-16
Fix offset for pushing and pulling a block

Joey79100:
New constants are +1 or -1 block +/- Lara's bounding box Z2 value at the very last frame of the pushing/pulling animation. Ideally it could even be fetched from the animation itself rather than typing it in the code.

Now the block follows Lara's hands as expected, stops when it should, and stops exactly at the center of the block at the end of Lara's pulling or pushing animation.

Should be used with attached animations below.
- Also includes fixed versions of the ungrab animations.
- Also includes possibility to ungrab earlier when Lara has just started grabbing the block.

[PUSHABLE_ANIMATIONS.zip](https://github.com/MontyTRC89/TombEngine/files/10251906/PUSHABLE_ANIMATIONS.zip)

### DIFF
--- a/TombEngine/Objects/TR5/Object/tr5_pushableblock.cpp
+++ b/TombEngine/Objects/TR5/Object/tr5_pushableblock.cpp
@@ -1,20 +1,20 @@
 #include "framework.h"
 #include "Objects/TR5/Object/tr5_pushableblock.h"
 
-#include "Game/Lara/lara.h"
-#include "Game/Lara/lara_helpers.h"
 #include "Game/animation.h"
 #include "Game/items.h"
-#include "Game/collision/collide_room.h"
 #include "Game/collision/collide_item.h"
+#include "Game/collision/collide_room.h"
 #include "Game/collision/floordata.h"
-#include "Game/control/flipeffect.h"
+#include "Game/Lara/lara.h"
+#include "Game/Lara/lara_helpers.h"
 #include "Game/control/box.h"
-#include "Specific/level.h"
-#include "Specific/Input/Input.h"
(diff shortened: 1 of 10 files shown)


Sezzary:
Additionally demagicked some numbers and dehardcoded a frame check.

TombEngine/Objects/TR5/Object/tr5_pushableblock.cpp
```diff
@@ -493,160 +508,145 @@ void PushableBlockCollision(short itemNumber, ItemInfo* laraItem, CollisionInfo*
...
 		PushableBlockBounds.BoundingBox.Z2 = 0;
 
-		short rot = pushableItem->Pose.Orientation.y;
+		short yOrient = pushableItem->Pose.Orientation.y;
 		pushableItem->Pose.Orientation.y = (laraItem->Pose.Orientation.y + ANGLE(45.0f)) & 0xC000;
 
 		if (TestLaraPosition(PushableBlockBounds, pushableItem, laraItem))
 		{
-			unsigned short quadrant = (unsigned short)((pushableItem->Pose.Orientation.y / 0x4000) + ((rot + 0x2000) / 0x4000));
+			int quadrant = int((pushableItem->Pose.Orientation.y / 0x4000) + ((yOrient + 0x2000) / 0x4000));
```
Lwmte, line 516: Why `GetQuadrant` is not used here? Or calculation is different, and just the name is the same?

Sezzary: The calculation was a little different here so I left it alone, but I've figured it out now!

PR #921, opened 2022-12-18
Fixed TR5 Cyborg

Nemoel-Tomo:
Fixed a crash with TriggerSparks, however I changed his electric sparks to that cool yellow dot shaped sparks again.
Fixed water death. 
Fixed meshswap. 
Now he switches to escape mood when the room is gassed. 
He is now invincible when Lara shoots at him. (there was a bug that he didn't do anything once he had 0 hp).


Lwmte (commented):
Cyborg needs to be tested again after @Sezzary cleanups, because some values were changed to constexpr, and `SetAnimation` was used in several places instead of direct set of state/anim numbers.

PR #926, opened 2022-12-23
Update cog_switch.cpp

Adngel:
- Fix an animation issue when Lara stopped using the cog switch

- Added a OCB condition, if OCB is 0, the cog switch will keep working as original one. (Requiring the link with a door that has a OCB2).

But if the OCB put on the cog switch, is different than 0, then it will be able to be used by Lara anytime (although it won't execute any trigger). Useful for Lua script works.

### DIFF
--- a/TombEngine/Objects/Generic/Switches/cog_switch.cpp
+++ b/TombEngine/Objects/Generic/Switches/cog_switch.cpp
@@ -149,7 +149,8 @@ namespace TEN::Entities::Switches
 		}
 		else
 		{
-			if (switchItem->Animation.FrameNumber == g_Level.Anims[switchItem->Animation.AnimNumber].frameEnd)
+			if ((switchItem->Animation.FrameNumber == g_Level.Anims[switchItem->Animation.AnimNumber].frameEnd)
+				&& (LaraItem->Animation.AnimNumber == LA_COGWHEEL_RELEASE))
 			{
 				switchItem->Animation.ActiveState = SWITCH_OFF;
 				switchItem->Status = ITEM_NOT_ACTIVE;
@@ -157,6 +158,12 @@ namespace TEN::Entities::Switches
 				RemoveActiveItem(itemNumber);
 
 				Lara.Control.HandStatus = HandStatus::Free;
+			}
+			else
+			{
+				//If Lara is repeating the PULL animation (because player dropped after the frame check).
+				//do the wheel animation again.
+				switchItem->Animation.TargetState = SWITCH_ON;
 			}
 		}
 	}
--- a/TombEngine/Objects/Generic/Switches/cog_switch.cpp
+++ b/TombEngine/Objects/Generic/Switches/cog_switch.cpp
@@ -60,7 +60,7 @@ namespace TEN::Entities::Switches
 
 		// Door was not found, do ordinary collision and exit.
 
-		if (door == nullptr)
+		if ((door == nullptr) && (switchItem->TriggerFlags == 0x00))
 		{
 			ObjectCollision(itemNum, laraItem, coll);
 			return;
@@ -96,7 +96,7 @@ namespace TEN::Entities::Switches
 						switchItem->Animation.TargetState = SWITCH_ON;
 						switchItem->Status = ITEM_ACTIVE;
 
-						if (door != NULL)
+						if ((door != NULL) && (switchItem->TriggerFlags == 0x00))
 						{
 							if (!door->opened)
 							{
@@ -139,8 +139,11 @@ namespace TEN::Entities::Switches
 			{
 				if (LaraItem->Animation.FrameNumber == g_Level.Anims[LaraItem->Animation.AnimNumber].frameBase + 10)
 				{
-					auto* doorItem = &g_Level.Items[Lara.InteractedItem];
-					doorItem->ItemFlags[0] = COG_DOOR_TURN;
+					if (switchItem->TriggerFlags == 0x00)
+					{
+						auto* doorItem = &g_Level.Items[Lara.InteractedItem];
+						doorItem->ItemFlags[0] = COG_DOOR_TURN;
+					}
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


TombEngine/Objects/Generic/Switches/cog_switch.cpp
```diff
@@ -60,7 +60,7 @@ namespace TEN::Entities::Switches
 
 		// Door was not found, do ordinary collision and exit.
 
-		if (door == nullptr)
+		if ((door == nullptr) && (switchItem->TriggerFlags == 0x00))
```
Lwmte, line 63: You can just use `!switchItem->TriggerFlags`. No need to use `== 0x00`.

TombEngine/Objects/Generic/Switches/cog_switch.cpp
```diff
@@ -96,7 +96,7 @@ namespace TEN::Entities::Switches
 						switchItem->Animation.TargetState = SWITCH_ON;
 						switchItem->Status = ITEM_ACTIVE;
 
-						if (door != NULL)
+						if ((door != NULL) && (switchItem->TriggerFlags == 0x00))
```
Lwmte, line 99: As above.

TombEngine/Objects/Generic/Switches/cog_switch.cpp
```diff
@@ -139,8 +139,11 @@ namespace TEN::Entities::Switches
 			{
 				if (LaraItem->Animation.FrameNumber == g_Level.Anims[LaraItem->Animation.AnimNumber].frameBase + 10)
 				{
-					auto* doorItem = &g_Level.Items[Lara.InteractedItem];
-					doorItem->ItemFlags[0] = COG_DOOR_TURN;
+					if (switchItem->TriggerFlags == 0x00)
```
Lwmte, line 142: As above.

TombEngine/Objects/Generic/Switches/cog_switch.cpp
```diff
@@ -153,8 +156,8 @@ namespace TEN::Entities::Switches
 
 				RemoveActiveItem(itemNumber);
 
-				LaraItem->Animation.AnimNumber = LA_STAND_SOLID;
-				LaraItem->Animation.FrameNumber = g_Level.Anims[LaraItem->Animation.AnimNumber].frameBase;
+				LaraItem->Animation.AnimNumber = LA_COGWHEEL_RELEASE;
+				LaraItem->Animation.FrameNumber = g_Level.Anims[LaraItem->Animation.AnimNumber].frameBase + 11;
```
Lwmte, line 160: Can you dehardcode frame number? Is it last frame or what?

TombEngine/Objects/Generic/Switches/cog_switch.cpp
Adngel: Sure, I'm changing it.

TombEngine/Objects/Generic/Switches/cog_switch.cpp
Adngel: Is not ideal.

In WadTool, the order to do the State change in PULL animation, is at 34-35, howere in game, it does the change later (around frame 44-45). So when it plays the RELEASE animation, it repeates several frames performing a glitching animation (Lara standing from lowest point two times), the offset is a way to subsane that glitch.

With a delay of 11 frames, the animation get smoother. (1-9 it get repeated frames, 10 gets a slight shake, with 12 -17 there are missing frames that produce an animation gap).

My doubt is, if in WadTool I order to do the State change at frame 34-35, why does TEN do the State change at 44-45? (Causing the repeated frames)

Lwmte: I don't know the reason, but we surely need to understand why it happens. Hardcoding frame offset to compensate for asset bug or another code bug elsewhere is a no go.

Adngel: Sleeping on it, I came with a clear mind and I saw what was happening, there were 2 problems:

1- It was activating the next animation by force (setting the animation directly) instead of leaving the system follow the flow with a state change.
2- The place where those lines were, only got activated when the cog wheel animation ended, that is what was causing the 11 frames delay. 

I've sent a new commitment, where I removed the lines related to Lara animation in the last segment, because Lara transition happens at frame 34, if I leave there, even if it's by the state flow, it only is detected at frame 45 making  Lara do the pulling animation one more time to can leave it at frame 34.

And up there, in the moment when the player release the action key, is when I send the order to change the state. So Lara can do the change at frame 34 and the cog can keep going to finish its animation before get deactivated.

Lwmte: So, everything works correctly now out of the box? No need for hardcoded frame numbers or asset updates?

Adngel: Yep. I'm using these ones and I'm not observing more problems.


It could be improved if new transitions animations were maded, the PULL animation last 54 frames (almost 2 seconds)

- 0-33 Lara is pulling down the wheel, is normal that she don't stop at that point, if player release action key, Lara will still continue to finish this round.
-  34-35 Lara is at the end of the pulling, this moment if action has been released, she will stop, or if not, she will prepare to do another pull round.
- 36-54 Lara is standing up but grabbing the wheel again, if player release action key during these frames, he still must see how Lara does 1 round more. But this is not a code issue. This could change with:

1. Adding more transitions animations (for example PULL frame 46 to IDLE) that would make possible to leave the PULLING animation at other points, not only frames 34-35.
2. Or increasing the check range, instead of 34-35 to a bigger extend, but although this would make it more reactive to players intentions, it would cause another animation glitch.

In any case, is not a code issue anymore. Both options (animations and checking range could be sort by the developer with current WadTool options).

Lwmte: Great to hear that! Maybe you should keep info about that in your notes or somewhere for future reference :)
I am merging your PR now.

PR #938, opened 2023-01-02
Trigger enhancements

hispidence:
Fixed `Moveable:Enable` and `Moveable:Disable` not doing the same as triggers and antitriggers. Disabling a door, for example, will now make the door close (if the door type supports it).

Antitriggers now do not kill enemies but instead disable them (i.e. triggering them again will re-enable them). This is so that enemies maintain the same "switched off and on" logic of doors/emitters/whatever that antitriggering usually has.

Added `Moveable:SetVisible`. In the past, `MakeInvisible` and `Disable` were both undone by `Enable`, but this proved too problematic, and getting it to work probably would have majorly overcomplicated the logic - so we now have `Enable` and `Disable` doing one thing, and `SetVisible(true)` and `SetVisible(false)` doing another.


TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
```diff
@@ -982,68 +989,37 @@ void Moveable::UnswapMesh(int meshId)
...
-		else
-		{
-			m_item->TouchBits = NO_JOINT_BITS;
-			AddActiveItem(m_num);
-			m_item->Status = ITEM_ACTIVE;
-		}
+	if (m_num == NO_ITEM)
+		return;
+
+	bool wasInvis = false;
```
Lwmte, line 999: Please rename this to `wasInvisible`.

Scripts/Engine/Timer.lua
```diff
@@ -120,12 +120,12 @@ Timer = {
 				t.remainingTime = t.remainingTime - dt
 
 				if t.remainingTime <= 0 then
-					t.func(table.unpack(t.funcArgs))
 					if not t.loop then
 						t.active = false
 					else
 						t.remainingTime = t.remainingTime + t.totalTime
 					end
+					t.func(table.unpack(t.funcArgs))
```
Lwmte, line 128: Why order was changed here? I am not saying it's a bug, just being curious :)

TombEngine/Game/items.cpp
```diff
@@ -441,9 +449,8 @@ void RemoveActiveItem(short itemNumber)
 			}
 		}
 
-		g_GameScriptEntities->NotifyKilled(&item);
-		if (!item.Callbacks.OnKilled.empty())
-			g_GameScript->ExecuteFunction(item.Callbacks.OnKilled, itemNumber);
+		if(killed)
```
Lwmte, line 452: Please add a space between `if` and `(`.

Scripts/Engine/Timer.lua
hispidence: It was a fix for https://github.com/MontyTRC89/TombEngine/issues/940; I forgot to switch to master before committing

PR #945, opened 2023-01-09
Debugging adngel

Adngel:
This branch includes:

### Bug fix
- Seth Blades: Fixed the double activation when used an OCB different than 0.
- Harpy: Fixed the sting attack which was not hurting nor poisoning Lara.
- Sphinx: Added the undead flag to prevent it become bugged after receiving a lot of damage.
- Skeleton: Fixed the creature spawn when used the OCB 3
- SAS: Update its zone from Human to Basic, making him stop teleporting over the blocks he walks by.

### Update
- Guide.cpp: Change the "path goal" store variable from **item->ItemFlags[4]** to **creature->LocationAI**
- Skeleton.cpp: Added the names of the states and animation lists. Also replaced the id numbers from the state cases with actual names.


TombEngine/Objects/TR4/Entity/tr4_skeleton.cpp
```diff
@@ -345,25 +379,25 @@ namespace TEN::Entities::TR4
...
 				creature->Flags = 0;
 
 				if (item->AIBits & GUARD ||
-					TestProbability(1.0f / 30) &&
+					Random::TestProbability(1 / 30.0f) &&
 					(AI.distance > pow(SECTOR(1), 2) ||
 						creature->Mood != MoodType::Attack))
 				{
-					if (TestProbability(0.0155f))
+					if (Random::TestProbability(0.0155f))
```
Lwmte, line 398: Can you also convert this TestProbability like the one below?

TombEngine/Objects/TR4/Entity/tr4_skeleton.cpp
```diff
@@ -415,22 +449,22 @@ namespace TEN::Entities::TR4
 						{
 							if (item->Animation.RequiredState)
 								item->Animation.TargetState = item->Animation.RequiredState;
-							else if (TestProbability(0.0155f))
+							else if (Random::TestProbability(0.0155f))
```
Lwmte, line 452: As above

TombEngine/Objects/TR4/Trap/tr4_sethblade.cpp
```diff
@@ -1,83 +1,86 @@
...
-		item->Animation.AnimNumber = Objects[item->ObjectNumber].animIndex + SETHBLADE_ANIM_IDLE;
-		item->Animation.TargetState = SETHBLADE_STATE_ACTIVE;
-		item->Animation.ActiveState = SETHBLADE_STATE_IDLE;
+		SetAnimation(&item, SETHBLADE_ANIM_IDLE);
+		item.ItemFlags[2] = abs(item.TriggerFlags); // NOTE: ItemFlags[2] stores blade timer.
+		item.ItemFlags[3] = SETH_BLADE_HARM_DAMAGE; // NOTE: ItemFlags[3] stored blade harm damage.
 
-		item->Animation.FrameNumber = g_Level.Anims[item->Animation.AnimNumber].frameBase;
-		item->ItemFlags[2] = abs(item->TriggerFlags);
+		// Immediately start blades.
```
Lwmte, line 35: I don't understand the purpose of this addition. First, you copy `abs(item.TriggerFlags)` to `item.ItemFlags[2]`, but here you immediately overwrite it with 1. Why?

TombEngine/Objects/TR4/tr4_objects.cpp
```diff
@@ -249,7 +249,7 @@ namespace TEN::Entities
 			obj->pivotLength = 50;
 			obj->radius = 102;
 			obj->intelligent = true;
-			obj->ZoneType = ZoneType::Human;
```
Lwmte, line 252: It's not necessary to comment it here. When animations / states will be added, then we can change zone type again. Otherwise, it's original behaviour and it doesn't require `TODO` comment here.

TombEngine/Objects/TR4/Trap/tr4_sethblade.cpp
Adngel: Basically.

If the OCB is 0, it goes to 1 to trigger the activation state.
If the OCB is >0, it was doing a waiting, then it does the activation, and repeat the waiting. Initializing the >0 cases, makes the blade to get activated when it's triggered, and then keeps repeating with the waiting of frames indicated in the OCB.

If the OCB is <0, this is the special case used in SethTomb where the blades do a wait of frames before to trigger, so here yes it's initialized with abs(OCB), (in editor is put in negative but we just want positive values for the counter).

Still, now you mentioned this, I though it could be better written in 1 line, I'm updatting it with the next commit.

PR #952, opened 2023-01-13
Fix Big Rat death crash

Adngel:
It was happening, that when the Big Rat was killed, like other objects, "CreatureAnimation( )" puts the monster in deactivate state and deletes its item->Data.

But the Control function was trying to access that item->Data after that. In the "IsRatOnWater( )" function, which must be activated after CreatureAnimation( ). Because item->Data didn't exist by then, the game crashes.

This fix what it does is add a check verification to IsRatOnWater in the area that uses creature so if it detects that item->Data is invalid, it ignores those lines.

### DIFF
--- a/TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
+++ b/TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
@@ -81,7 +81,7 @@ namespace TEN::Entities::Creatures::TR1
 			SetAnimation(item, BIG_RAT_ANIM_IDLE);
 	}
 
-	bool IsRatOnWater(ItemInfo* item)
+	bool RatOnWater(ItemInfo* item)
 	{
 		int waterDepth = GetWaterSurface(item->Pose.Position.x, item->Pose.Position.y, item->Pose.Position.z, item->RoomNumber);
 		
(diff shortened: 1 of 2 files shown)


TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -83,21 +83,24 @@ namespace TEN::Entities::Creatures::TR1
...
+		if (item->IsCreature())
 		{
-			creature->LOT.Step = CLICK(1);
-			creature->LOT.Drop = -CLICK(1);
-		}
+			auto* creature = GetCreatureInfo(item);
 
+			if (waterDepth != NO_HEIGHT)
+			{
+				creature->LOT.Step = BLOCK(20);
```
Lwmte, line 94: This is not related to PR itself, but it greatly confuses me that function which is supposed to check if rat is on water, also modifies creature LOT. I would rename `IsRatOnWater` to just `RatOnWater`.

Adngel: That's what allows it to climb up and down the blocks right? 

I can change the name in a quick commitment.

Adngel:
I forgot update the changes.txt, I've dont the commit now.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
Lwmte: Yes please :)

Adngel: Yes, I've already done.

Lwmte (approved):

PR #955, opened 2023-01-14
Add fixed shadow and dropped pickup positions

Lwmte:
* Fix enemy shadow position.
* Make enemies drop pickups at their origin point, not bounding box centerpoint.


Sezzary (approved):
Looks good. Now we let people test.

Adngel:
I've tested this branch, and I did not see any meaningful issue.

I saw some anomalies of low importance through that may be related or not to the task of this branch.

- When the TR5 Hydra dies, its blob shadow remains in the place. (Hidra code bug?)
- When the Big_Scorpion kills a troop, the troop's drop appears under the scorpion. (Animation asset setup?)
- When the Small_Scorpion is killed in a 1 click space, it can drops objects there, where Lara can't reach. (Up to the Builder avoid that case)

From my point of view, this could get the green light.

PR #957, opened 2023-01-15
Fix tr5 roman statue

Nemoel-Tomo:


TombEngine/Objects/TR5/Entity/tr5_roman_statue.cpp
```diff
@@ -306,37 +278,47 @@ namespace TEN::Entities::Creatures::TR5
...
 		{
 			ExplodeItemNode(item, 8, 0, 8);
-			item->MeshBits |= 0x100;
-			item->SetMeshSwapFlags(0x100);
+			jointBit = MS_HARD_DMG;
+			item->MeshBits |= jointBit;
+			item->ItemFlags[1] |= jointBit;
+			item->SetMeshSwapFlags(item->ItemFlags[1]);		
 		}
-		else if (item->HitPoints < 150 && !item->TestMeshSwapFlags(0x400))
```
Lwmte, line 328: I am not sure why `TestMeshSwapFlags` was reverted here and further. It's basically the same test?

TombEngine/Objects/TR5/Entity/tr5_roman_statue.cpp
```diff
@@ -374,6 +356,12 @@ namespace TEN::Entities::Creatures::TR5
 			int deltaFrame;
 			bool unknown;
 
+			int R = 0;
```
Lwmte, line 359: Why this block with `R, G, B, color` was added? It seems only `R` is used further, and it is 0 anyway, so why not pass 0 directly?

Lwmte (changes requested):

TombEngine/Objects/TR5/Entity/tr5_roman_statue.cpp
Nemoel-Tomo: changed

TombEngine/Objects/TR5/Entity/tr5_roman_statue.cpp
Nemoel-Tomo: fixed

Adngel:
Hi there, I've tested the branch and I saw these next issues:

- **Sword guy, gets interrupted during its walking animation loop**. Sometime he stops to shoot, but other times he stops for no reason before to continue walking. May a condition be causing him to go out of the walking state?
- **Both can walk over blocks of 2 and 4 clicks**, I don't know if is a bug or is intended. Usually as bosses I think their battlegrounds will be mostly flats, but maybe can be solved changing the zone they use to the basic one?
- **1 supegrenade is enough to kill them from the start**. Are supergrenades too powered or are roman gods too weak?

But generally I did not see or got any meanigful problem, all green light from me.

Adngel:
I've tested and I've observed that the easy kill with supergrenade, the pathfinding zone and the super grenade crashes have been solved, the do not happens anymore in my tests.

TombEngine/Game/effects/spark.cpp
```diff
@@ -172,6 +172,10 @@ namespace TEN::Effects::Spark
 		spark.friction = 34;
 		spark.maxYvel = 0;
 		spark.gravity = 0;
+		spark.scalar = 2;
```
Lwmte, line 175: This makes all three values the same (GenerateInt is called once, not thrice). Is this intended?

Nemoel-Tomo: yes, because some of the sparks would look ugly because they would be stretched then

Sezzary (approved):

Lwmte (approved):

PR #961, opened 2023-01-16
Save vec3 rotation color

hispidence:
Make Rotation and Color/ScriptColor objects saveable via LevelVars/GameVars.

Added LevelFuncs.PreSave, a built-in function that gets called before data is saved (since LevelFuncs.OnSave happens after saving data).


TombEngine/Game/gui.cpp
```diff
@@ -3146,6 +3146,7 @@ namespace TEN::Gui
 		if (GuiIsSelected())
 		{
 			SoundEffect(SFX_TR4_MENU_CHOOSE, nullptr, SoundEnvironment::Always);
+			g_GameScript->PreSave();
```
Lwmte, line 3149: Why do we need `PreSave` anyway? Unless you provide a specific use case, I think we should not multiply events. To my knowledge, nobody requested for this feature. And even if we decide to keep it, I think that naming is inconsistent. Event callback names should start with `On` prefix, otherwise it looks not like callback, but like event itself.

hispidence: It's useful because the builder might want to save specific data only when saving the game. They can't use `OnSave`, as that will take place AFTER data is saved, so if they save, quit, and reload, then the stuff that was saved in `OnSave` will be gone. I added it because it made testing and debugging the other changes in this PR much easier.

What name would you suggest? `OnPreSave`?

Lwmte: I think the `OnPreSave` name is fine, but let's now think about `OnSave`. Is there also any specific use case when an event should be called AFTER save? I mean, probably `OnSave` itself should be called before actual save, which makes more sense?

hispidence: I considered that, yeah. If we were to only keep one of these then a pre-save callback would definitely be more useful. I don't know if I can think of any use cases for a post-save one... so perhaps we keep `OnSave` and call it before the save (and make it clear in the docs that this is the case)? I'd be okay with that.

Lwmte: Yes, I think this will be the best solution for now. If ever any builder will need post-save event, we can add it later.

PR #962, opened 2023-01-19
Switches adngel

Adngel:
This branch includes.

- Fixes for switches
- Added of the Lock_Switch (for valves attached to doors)
- Added flipeffect 47 to activate ID_LARA animations from ID_LARA_EXTRA_ANIM

It still required asset fixes for it to work.

ID OFF is 1
ID ON is 0

So Big Buttons, Valve wheels and Wall Holes, animation states must be edited to follow these rules (that follows most of the switches in the TR games).


Documentation/Changes.txt
```diff
@@ -15,11 +15,14 @@ Version 1.0.6
 * Fix TR4 SKELETON spawn when used with OCB 3
 * Fix TR4 SAS teleporting over the blocks he walks by.
 * Fix enemy projectile effect colours.
+* Fix switches with ocb 3, 5 and 6 that failed to work in some circunstances.
```
Lwmte, line 18: "OCB" should be capitalized, and "circumstances", not "circunstances".

TombEngine/Game/control/flipeffect.cpp
```diff
@@ -78,7 +78,8 @@ function<EffectFunction> effect_routines[NUM_FLIPEFFECTS] =
 	MeshSwapToPour,				//43
 	MeshSwapFromPour,			//44
 	LaraLocationPad,			//45
-	KillActiveBaddys			//46
+	KillActiveBaddys,			//46
+	ActiveNextAnimOnLaraSlot	//47
```
Lwmte, line 82: Why new flipeffect ID was added? There should be empty flipeffect IDs to be reused. Also, what was the initial purpose of adding this flipeffect at all? It seems it has a very narrow usage, and its meaning may be obscure to a builder?

TombEngine/Game/control/trigger.cpp
```diff
@@ -152,6 +151,10 @@ int SwitchTrigger(short itemNumber, short timer)
 	}
 	else if (item->Status)
 	{
+		if (item->ObjectNumber == ID_LOCK_SWITCH &&
```
Lwmte, line 154: I don't understand the purpose of this code. Can you provide an example how it is used (video + TE setup screenshot + description)?

Documentation/Changes.txt
Adngel: Done! :D

TombEngine/Game/control/trigger.cpp
Adngel: Maybe look more fussy explained than done.

The object can be taken from Tomb Raider 5, Joby 3 (submarine level), there are already with name AIRLOCK_SWITCH copied to TEN is also copied in slot 820 also named AIRLOCK_SWITCH.
image text

This is the image of the setup in editor:
image text

You put the lock and the door in the same sector, and put triggers for both in the same sector too. (For the lock, the trigger must be SWITCH type). No OCB are required.

This is how it works in game:
[Lock Switch Text](https://www.youtube.com/watch?v=0i7i8ehkIlg)


The segment of code you marked is to activate the trigger. The normal switches when activate, they do their animation and then get disabled. When disconnected, is when they activate their trigger. (That's why anicommand "Kill Entity" is required in the transition animations in normal switches).

However, the valve on the door has two animations. (The Lara interaction and the switch opening with the door). Originally the asset hasn't got any "Kill Entity" so it doesn't trigger anything with switches code.

- If put "Kill Entity" in first animation, it does trigger after the Lara's interaction animation, but in TEN, that "disable" the object, so although it triggers the door, it doesn´t perform the next switch animation.
- If put the "Kill Entity" in the second animation, (door opening animation), the switch does the two animations, but only do the trigger at the end of two animations, what is too late.

That code is overriding that step, doing the check to activate trigger while are in STATE 1. (first frame of Animation 2).

Adittionally I added its own control and collision function, although are very similar to generic switches, I removed many lines that were not used in this escenerio, besides, generic switches IDs are OFF 1, ON 0, but for this door lock are OFF 0, ON 1 So instead to make another exception like the Jump Switch, I went with its own code what makes it more autonomous.

TombEngine/Game/control/flipeffect.cpp
Adngel: Is a problem that I've got. 

Other partners have used animations from ID_LARA_EXTRA_ANIMS, but these were death animations, so they didn't need to return Lara's normal anims. However in my case, I tried get those animations for a custom switch, after this animation ends, I needed to return animation 11 of the ID_LARA. 

This is the example (done with OCB 7 and lua code).
[TNT like switch](https://www.youtube.com/watch?v=2FpvW2gjeSc)

It ocurred me multiples ways to do that return to ID_LARA, but I though that adding the flipeffect could be the less intrussive, and more flexible.

Still my question about this remains, if there were another method in TEN designed to deal with this problem, I would be glad to adapt it.

(Also know that this is not an unique case, eventually if I rework puzzles, the same problem with come with the TR2 gong or the TR1 Midas hand that is used to turn silver pickups into gold puzzles items, and when the others do the dragon, again will have it with the dagger's animation, and the Lara's TR2 wake up animation, plus any new animation that may get added in the future for who knows what?).

Adngel:
I'm afraid the switches branch may got broken with the new update.
I'll remove the PR and remake the branch tomorrow

PR #967, opened 2023-01-26
More lua funcs adngel

Adngel:
In this PR:

- Solved an issue that Camera:SetPosition did not update the camera position at the moment if this one was being played in that moment. (It was needed to re-activate the trigger or move Lara in order to make the camera refreshes).
- Added a Get Set functions for the IsAirborne parametter, what allows some new actions in lua that were nor possible before.
- Added a Get Set function for the creature->LocationAI what will allow to change the behaviours of some enemies.


TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
```diff
@@ -601,6 +605,22 @@ void Moveable::SetRot(Rotation const& rot)
 	m_item->Pose.Orientation.z = ANGLE(rot.z);
 }
 
+/// Get the moveable's airborne status
+// @function Moveable:GetAirborne
+// @treturn (bool) true if entity animation is flying.
```
Kubsy, line 610: I feel like this would be better worded as: "@return (bool) true if the entity is not grounded" or "@return (bool) true if the entity is in air"

but not sure, I would consult @Sezzary with English stuff :D

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
```diff
@@ -601,6 +605,22 @@ void Moveable::SetRot(Rotation const& rot)
...
+bool Moveable::GetAirborne() const
+{
+	return m_item->Animation.IsAirborne;
+}
+
+/// Set the moveable's airborne status
+// @function Moveable:SetAirborne
+// @tparam (bool) airborne status for the current animation.
```
Kubsy, line 618: Same as above, I feel like this would be better worded as: "@return (bool) true if the entity is grounded" or "@return (bool) true if the entity is not in air"

but not sure, I would consult @Sezzary with English stuff :D

TombEngine/Game/camera.cpp
```diff
@@ -405,6 +405,16 @@ void MoveObjCamera(GameVector* ideal, ItemInfo* camSlotId, int camMeshId, ItemIn
 	}
 }
 
+void FixedCameraRefresh(short camNumber)
```
Kubsy, line 408: Maybe rename function to just "CameraRefresh"?

Adngel: Not this case, I put the "Fixed" because I wanted make a distinction that this is updated for the fixed cameras.

levelCameras (the objects you put in the level) has some data that the Main Camera reads to positionate (and do its fixedCamera rutine), but it needs that levelCamera object. The other cameras rutines (chase, look, combat, etc) do not need those levelCameras, so this Refresh function actually do nothing to them.

Is, for what it's, to force a camera refresh in the FixedCamera rutine.

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
Adngel: I'm neither very convinced with "if flying", but IsAirborne is not exactly the grounded status.

Is a flag that creatures have, that indicate that status is "flying" (jumping floating, etc), so the physics tend to react accordingly. (like the gravity management).

However, is not a physical calculation of the grounded status. (For example, It gets activate during Monkey swing, although Lara is not on ground, she is not being affected by gravity).

I could write it in better way, although "not grounded" or "in air" doesn't fit either. @_@

Adngel: At the end I've updated with:

```
/// Get the moveable's airborne status
// @function Moveable:GetAirborne
// @treturn (bool) true if Lara state must react to aerial forces.
```
It may be the gravity, but other times may be also a force towards up like springboards. Or an backwards ejection like my case when she collides with the magic barrier. So... "aerial forces"?

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
Adngel: I may have not breaken my head a lot with this one...

```
/// Set the moveable's airborne status
// @function Moveable:SetAirborne
// @tparam (bool) New airborn status for Lara.
```

TombEngine/Game/camera.cpp
Joey79100: If you're going to change the function's name, it should be "RefreshFixedCamera", so it follows the "DoSomething" kind of naming which is already used, and more natural for function names, rather than something along the lines of "SomeAction".

Adngel: Yeah, that can be done.

Adngel: Done

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
Kubsy: yeah should be good

TombEngine/Game/camera.cpp
```diff
@@ -405,6 +405,16 @@ void MoveObjCamera(GameVector* ideal, ItemInfo* camSlotId, int camMeshId, ItemIn
 	}
 }
 
+void RefreshFixedCamera(short camNumber)
+{
+	auto* camera = &g_Level.Cameras[camNumber];
+
+	GameVector from = GameVector(camera->Position, camera->RoomNumber);
+	int moveSpeed = camera->Speed * 8 + 1;
```
Kubsy, line 413: Demagic numbers?

Adngel: I'm not sure how.

The other line where this happens (line 874) had this comment:

> // Multiply original speed by 8 to comply with original bitshifted speed from TR1-2

But I'm not very sure how the speed is stored in the camera (and is even possible stoe speed in the camera with TEN? in Tomb Editor that field appears disabled in grey to me).

I let it like that to ensure I don't break anything. :(

Sezzary: For now you could define a constant at the top of the function:
`constexpr auto SPEED_COEFF = 8;`

TombEngine/Game/camera.cpp
```diff
@@ -405,6 +405,16 @@ void MoveObjCamera(GameVector* ideal, ItemInfo* camSlotId, int camMeshId, ItemIn
 	}
 }
 
+void RefreshFixedCamera(short camNumber)
+{
+	auto* camera = &g_Level.Cameras[camNumber];
+
+	GameVector from = GameVector(camera->Position, camera->RoomNumber);
```
Sezzary, line 412: You can use `auto` here. Should also be named `origin` like everywhere else.

Adngel: done

TombEngine/Game/camera.cpp
```diff
@@ -405,6 +405,16 @@ void MoveObjCamera(GameVector* ideal, ItemInfo* camSlotId, int camMeshId, ItemIn
 	}
 }
 
+void RefreshFixedCamera(short camNumber)
+{
+	auto* camera = &g_Level.Cameras[camNumber];
```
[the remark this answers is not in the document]
Adngel: done

TombEngine/Game/camera.cpp
Adngel: The 8 itself is not a magic number nor a variable that can be placed as anyone likes. Is mechanical. What it's doing is bitshifting the camera speed 3 bits to the left.  

(So if TE camera.speed is 00000110 = 6  * 8 = 48 = 00110000)

What I don't understand is why TE camera has a value that needs to be displaced like this to be used.

Sezzary (approved):

PR #970, opened 2023-01-27
Squash and stretch bubbles

Sezzary:
"Very important"

- Bubbles oscillate (squash and stretch) as they ascend. Small bubbles oscillate rapidly, larger bubbles oscillate more slowly.
- A cone of bubbles is spawned when the player dives into water.


TombEngine/Game/effects/lightning.cpp
```diff
@@ -2,7 +2,7 @@
 #include "Game/effects/lightning.h"
 
 #include "Game/animation.h"
-#include "Game/effects/bubble.h"
+#include "Game/effects/Bubble.h"
```
Adngel, line 5: I see Bubble class was already included in the lightning.cpp effect. But does it really need it?. I can't find where it's using it.

Maybe wherever it was being used, was replaced by particles effects or something else?.

TombEngine/Objects/TR5/Emitter/tr5_smoke_emitter.cpp
```diff
@@ -83,14 +83,15 @@ void SmokeEmitterControl(short itemNumber)
...
+					(GetRandomControl() & 0x3F) + item->Pose.Position.x - 32,
+					item->Pose.Position.y - (GetRandomControl() & 0x1F) - 16,
+					(GetRandomControl() & 0x3F) + item->Pose.Position.z - 32);
 
 				if (item->TriggerFlags == 1)
-					CreateBubble(&pos, item->RoomNumber, 15, 15, 0, 0, 0, 0);
+					SpawnBubble(pos, item->RoomNumber, 0);
 				else
-					CreateBubble(&pos, item->RoomNumber, 8, 7, 0, 0, 0, 0);
+					SpawnBubble(pos, item->RoomNumber, 0);
```
Adngel, line 94: I'm misunderstanding these.

So Smoke emitter when placed underwater can have OCB 1 or any other (OCB 0 for example). What should be the difference?

In the old function, I see it's passing different parameters as arguments, but I'm not sure how the new SpawnBubble function is recreating one or the second. 

(And if it happens that they are really the same, couldn't ignore the OCB and leave just 1 call?).

Adngel (commented):
Still pending on tests, but I saw nothing bad.

Adngel:
BubblesTest

Some notes I took:

- Bubbles oscilate too fast, maybe slowing its animation could look better?
- When bubbles reaches surface, small bubbles create circular waves of the same size than big bubbles. Maybe could scale wave sizes too?
- Clump bubbles looks artificial, Maybe add rotation to them could make them look more dynamic?
- Lara Harpoon shoots create some cluster bubbles, but the effect doesn't look very nice.
- Rubber Boat create few bubbles, maybe it would look better with more?
- I saw that White_Smoke_Emitter with OCB 1, creates a continium bubbles emission, while OCB 0 emit the bubbles in pulses (random pulses, some times it creates few ones before pause, other times it creates long sequence)
- SCUBA_DIVER enemy's harpoon are rendered in black, and do not emit any bubbles. 
- Lara's flares when are in water (holded by Lara, or throwed), they neither emit any bubble effect.

TombEngine/Objects/TR5/Emitter/tr5_smoke_emitter.cpp
Adngel: Later in the test I checked.

I used SMOKE_EMITTER_WHITE but I saw it was using also the same Control function.

- Smoke_Emitter_White OCB 1, creates a continium bubbles emission.
- Smoke_Emitter_White OCB 0, emits the bubbles in pulses.

TombEngine/Game/effects/lightning.cpp
Sezzary: lightning.cpp has several pointless includes. Will remove this one!

TombEngine/Objects/TR5/Emitter/tr5_smoke_emitter.cpp
Sezzary: Hmm, I just noticed these are the same now. The removed arguments were both unused and unknown in the original decompiled version. I'll check tomb4/5 to see if they're supposed to do anything.

In your later test, is that behaviour observed with this code or with the previous?

Kubsy (approved):
Looks good to me. I tested all scenarios and bubbles appear. I wasn't able to these white smoke emitter because it was not appearing underwater (I am not sure how it works anyway).

TombEngine/Objects/TR5/Emitter/tr5_smoke_emitter.cpp
Sezzary: The extra arguments don't seem to be anything important, so I'll leave the logic here as is.

Sezzary:
Changes since adngel's review:
- Bubble squash and stretch speed has been slowed down a lottle.
- A ripples created by a bubble is proportional to its scale.
- A bubble's buoyancy is proportional to its scale.
- Clumb bubbles are no more.
- Harpoon bubble spawning has been adjusted.
- Rubble boats spawn more bubles.
- Flare chaff bubbles have been fixed.

PR #977, opened 2023-01-29
Draw holsters for shadow and bypass skin / braid / joints if they don't exist

Lwmte:
This pull request fixes 4 things:

1. Nanaki reported that his level crashes on start. It was caused by hair object missing in his level. For some reason, it worked in previous builds but now it silently crashes the game. This PR makes rendering of hair object optional, only if it actually exists in level file.
2. Same goes for joints - if they don't exist, they won't be rendered now, without crash.
3. If Lara skin object does not exist in level (e.g. if builder uses Lara model from TR1 or TR2), original Lara object meshes will be used.
4. Additionally, a side effect of first 2 fixes, now holsters are also rendered for a shadow map.


Kubsy (approved):

Adngel (approved):
I checked the code, I saw nothing wrong with the changes.

PR #983, opened 2023-01-31
Tiger polishment adngel

Adngel:
This branch make some light changes to Tigers and Lions to improve their behaviour.

## Tiger
- Changed the ranges of the melee attacks, so now it does both bites animation (looks like a combo), although if it gets Lara against a wall, may do only the quick one. Or if lara is a bit far, he may do the jumping one.
- When the tiger is running but Lara is stopped, I added a probability condition, to incentivate the swipe attack over the melee bites.
- If while running, Tiger decides he won't make the melee attack, now there is another condition to check that Lara is ahead before do the swipe attack, (to reduce the number of missdirected attacks), so if lara is not in front, tiger will keep running.
- Separated the melee attacks states from the swipe attack state. (Because some are done with mouth meshes, and the other with leg meshes)
- I added condition check for the animation and frame to avoid hurting Lara while the monster is preparing the attack.
- Added function to align entity to sloped floors.

## Lion
- Added a minimal range condition for the attacks, to reduce the number of times that the Lion attacks forward while Lara is behind Lion's head.
- Added another attack range for the other bite attack.
- Changed the attack collision check condition, for a distance range condition.
- Defined constants for turn values and added turn value to the attacks states to reduced missdirected attacks.
- Replaced AI.ahead for a custom targetAhead that consider the offset for the head. To reduce the times the lion attacks forward while Lara is on the side instead of on the front.
- Added creature->Flags value to control the number of times that DoDamage function happens.
- Added function to align entity to sloped floors.
- Reorganised code to make follow similar patterns that other creatures.

### Video
[Tiger and Lion test](https://www.youtube.com/watch?v=oR3-RlqPk5I)

### DIFF
--- a/TombEngine/Game/collision/collide_room.cpp
+++ b/TombEngine/Game/collision/collide_room.cpp
@@ -847,19 +847,19 @@ void AlignEntityToSurface(ItemInfo* item, const Vector2& ellipse, float alpha, f
 
 	// Calculate extra rotation required.
 	auto extraRot = EulerAngles(
-		FROM_RAD(atan2f(forwardHeightDif, ellipse.y * 2)),
+		FROM_RAD(atan2(forwardHeightDif, ellipse.y * 2)),
 		0,
-		FROM_RAD(atan2(lateralHeightDif, ellipse.x * 2))
-	) - EulerAngles(item->Pose.Orientation.x, 0, item->Pose.Orientation.z);
+		FROM_RAD(atan2(lateralHeightDif, ellipse.x * 2))) -
+		EulerAngles(item->Pose.Orientation.x, 0, item->Pose.Orientation.z);
 
-	// Rotate X axis if forward height difference is not too significant.
+	// Rotate X axis.
 	if (abs(forwardHeightDif) <= STEPUP_HEIGHT)
 	{
 		if (abs(extraRot.x) <= ANGLE(constraintAngle))
 			item->Pose.Orientation.x += extraRot.x * alpha;
 	}
 
-	// Rotate Z axis if lateral height difference is not too significant.
+	// Rotate Z axis.
 	if (abs(lateralHeightDif) <= STEPUP_HEIGHT)
 	{
 		if (abs(extraRot.z) <= ANGLE(constraintAngle))
@@ -869,7 +869,7 @@ void AlignEntityToSurface(ItemInfo* item, const Vector2& ellipse, float alpha, f
 
 int GetQuadrant(short angle)
 {
-	return (unsigned short)(angle + ANGLE(45.0f)) / ANGLE(90.0f);
+	return (unsigned short(angle + ANGLE(45.0f)) / ANGLE(90.0f));
 }
 
 // Determines vertical surfaces and gets nearest ledge angle.
--- a/TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
+++ b/TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
@@ -187,10 +187,10 @@ namespace TEN::Entities::Creatures::TR3
 				case TIGER_STATE_POUNCE_ATTACK:
 					if (!creature->Flags && item->TouchBits.Test(TigerBiteAttackJoints))
 					{
-						if	(item->Animation.AnimNumber == Objects[ID_TIGER].animIndex + TIGER_ANIM_BITE_ATTACK &&
-							 item->Animation.FrameNumber > g_Level.Anims[item->Animation.AnimNumber].frameBase + 4 ||
-							item->Animation.AnimNumber == Objects[ID_TIGER].animIndex + TIGER_ANIM_POUNCE_ATTACK_START &&
-							 item->Animation.FrameNumber > g_Level.Anims[item->Animation.AnimNumber].frameBase + 12 )
+						if	(item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_BITE_ATTACK) &&
+							 item->Animation.FrameNumber > GetFrameNumber(item, 4) ||
+							 item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_POUNCE_ATTACK_START) &&
+							 item->Animation.FrameNumber > GetFrameNumber(item, 12))
 						{
 							DoDamage(creature->Enemy, TIGER_BITE_ATTACK_DAMAGE);
 							CreatureEffect(item, TigerBite, DoBloodSplat);
@@ -202,9 +202,9 @@ namespace TEN::Entities::Creatures::TR3
 				case TIGER_STATE_RUN_SWIPE_ATTACK:
 					if (!creature->Flags && item->TouchBits.Test(TigerSwipeAttackJoints))
 					{
-						if (item->Animation.AnimNumber == Objects[ID_TIGER].animIndex + TIGER_ANIM_RUN_SWIPE_ATTACK &&
-							item->Animation.FrameNumber >= g_Level.Anims[item->Animation.AnimNumber].frameBase + 6 &&
-							item->Animation.FrameNumber < g_Level.Anims[item->Animation.AnimNumber].frameBase + 16)
+						if (item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_RUN_SWIPE_ATTACK) &&
+							item->Animation.FrameNumber >= GetFrameNumber(item, 6) &&
+							item->Animation.FrameNumber < GetFrameNumber(item, 16))
 						{
 							DoDamage(creature->Enemy, TIGER_SWIPE_ATTACK_DAMAGE);
 							CreatureEffect(item, TigerBite, DoBloodSplat);
--- a/TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
+++ b/TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
@@ -210,10 +210,11 @@ namespace TEN::Entities::Creatures::TR3
 				case TIGER_STATE_POUNCE_ATTACK:
 					if (!creature->Flags && item->TouchBits.Test(TigerBiteAttackJoints))
 					{
-						if	(item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_BITE_ATTACK) &&
-							 item->Animation.FrameNumber > GetFrameNumber(item, 4) ||
-							 item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_POUNCE_ATTACK_START) &&
-							 item->Animation.FrameNumber > GetFrameNumber(item, 12))
+						if	( (item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_BITE_ATTACK) &&
+							   item->Animation.FrameNumber > GetFrameNumber(item,  4) ) ||
+							  (item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_POUNCE_ATTACK_START) &&
+							   item->Animation.FrameNumber > GetFrameNumber(item, 12) )
+							)
 						{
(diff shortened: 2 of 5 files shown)


TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
```diff
@@ -99,100 +105,122 @@ namespace TEN::Entities::Creatures::TR3
...
+					break;
 			}
 		}
 
 		CreatureTilt(item, tiltAngle);
 		CreatureJoint(item, 0, extraHeadRot.y);
 		CreatureAnimation(itemNumber, headingAngle, tiltAngle);
+
+		auto radius = Vector2(object->radius, object->radius * 1.33f);
+		AlignEntityToSurface(item, radius);
```
Kubsy, line 224: this alignes the enemy to surface, Right? For instance, it will align according to the slope?

TokyoSU: Yes it align the entity to any slope, the most example could be the crocodile from TR4 which use it.

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
```diff
@@ -99,100 +105,122 @@ namespace TEN::Entities::Creatures::TR3
...
+					}
+
+					creature->Flags = 0;
+					break;
+
+				case TIGER_STATE_BITE_ATTACK:
+				case TIGER_STATE_POUNCE_ATTACK:
+					if (!creature->Flags && item->TouchBits.Test(TigerBiteAttackJoints))
+					{
+						if	(item->Animation.AnimNumber == Objects[ID_TIGER].animIndex + TIGER_ANIM_BITE_ATTACK &&
```
TokyoSU, line 190: There you should use GetAnimNumber instead which was added with sophia PR :)
Also use GetFrameNumber too

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
```diff
@@ -99,100 +105,122 @@ namespace TEN::Entities::Creatures::TR3
...
+							CreatureEffect(item, TigerBite, DoBloodSplat);
+							creature->Flags = 1; // 1 = is attacking.
+						}
+					}
+					break;
+
+				case TIGER_STATE_RUN_SWIPE_ATTACK:
+					if (!creature->Flags && item->TouchBits.Test(TigerSwipeAttackJoints))
+					{
+						if (item->Animation.AnimNumber == Objects[ID_TIGER].animIndex + TIGER_ANIM_RUN_SWIPE_ATTACK &&
```
TokyoSU, line 205: Same as above.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
-
-						if (AI.distance < LION_POUNCE_ATTACK_RANGE)
+						if (AI.distance >= LION_ATTACK_MIN_RANGE)
 						{
-							item->Animation.TargetState = LION_STATE_POUNCE_ATTACK;
-							break;
+							if (AI.distance <= LION_BITE_ATTACK_RANGE)
+							{
+								item->Animation.TargetState = LION_STATE_BITE_ATTACK;
+								break;
```
TokyoSU, line 152: you can remove this break;

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
+							}
+							if (AI.distance < LION_POUNCE_ATTACK_RANGE)
```
TokyoSU, line 154: Then add else if here

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
+							{
+								item->Animation.TargetState = LION_STATE_POUNCE_ATTACK;
+								break;
```
TokyoSU, line 157: And remove this break.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
+					creature->MaxTurn = LION_WALK_TURN_RATE_MAX;
 
 					if (creature->Mood == MoodType::Bored)
 					{
-						if (TestProbability(1.0f / 256))
-						{
-							item->Animation.TargetState = LION_STATE_IDLE;
-							item->Animation.RequiredState = LION_STATE_ROAR;
-						}
+						if (TestProbability(LION_ROAR_CHANCE))
```
TokyoSU, line 170: Move this to lign 168 `
if (creature->Mood == MoodType::Bored && TestProbability(LION_ROAR_CHANCE)) {
item->Animation.TargetState = LION_STATE_ROAR;
}
`

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
 
-					break;
+				break;
 
 				case LION_STATE_POUNCE_ATTACK:
-					if (!item->Animation.RequiredState &&
+					creature->MaxTurn = LION_ATTACKS_TURN_RATE_MAX;
+
+					if (!creature->Flags &&
+						item->Animation.AnimNumber == Objects[ID_LION].animIndex + LION_ANIM_POUNCE_ATTACK_END &&
```
TokyoSU, line 204: same as above for GetAnimNumber

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
```diff
@@ -118,94 +142,98 @@ namespace TEN::Entities::Creatures::TR5
...
-					break;
+				break;
 
 				case LION_STATE_BITE_ATTACK:
-					creature->MaxTurn = ANGLE(1.0f);
+					creature->MaxTurn = LION_ATTACKS_TURN_RATE_MAX;
 
-					if (!item->Animation.RequiredState &&
+					if (!creature->Flags &&
+						item->Animation.AnimNumber == Objects[ID_LION].animIndex + LION_ANIM_BITE_ATTACK &&
```
TokyoSU, line 218: same as above for GetAnimNumber

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
Adngel: Yes, is the function that Sezz made to fix the Big_Scorpion alignment, I though we could keep reusing it for other entities.

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
Adngel: Ohm, that will be awesome and useful (I didn't like this method of "start point + animation number" >.<).
I'll update.

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
Kubsy: okie :)

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: As it's right now, no I can't.

Because if I quit that break, the thread will continue and will replace TargetState with the running state of line 162.

This lion, although code is very similar to the other entities of TEN, still has a change in the pattern.

The other creatures do

```
Behavior case: IDLE
   if (this)
     foo1
   else if (this 2)
      foo2
   else if (this 3)
      foo3
   else
      foo4
```
In that case I would omit the middle breaks because I know when any of the foo happens, the rest of the code won't happen.

But lion does

```
Behaviour case IDLE
 if (this)
     foo1 
 if (this2)
      foo2
 if (this3)
      foo3
foo4

```
Here I can't omit because if a foo happens, that's don't garanteed the next ones don't happen.


I though in re-write all the lion code to make it like the other entities, but I though it could be any fixation of mine, so I respected the structure no "else if" as wrote... whoever wrote it. O_ò

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: Yeah, what I was answering in the previous comment. Let's keep the conversation up there.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: idem as before.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: Ohm true, I didn't notice that.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: done

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: done

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
Adngel: done

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
Adngel: done

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
Adngel: Done. I'll update also the harpy where I used also the old aimIndex method for the stinger attack.

TombEngine/Objects/TR5/Entity/tr5_lion.cpp
TokyoSU: Okay no problem then

TokyoSU (approved):

TombEngine/Objects/TR3/Entity/tr3_tiger.cpp
```diff
@@ -82,117 +88,162 @@ namespace TEN::Entities::Creatures::TR3
...
+					}
+
+					creature->Flags = 0;
+					break;
+
+				case TIGER_STATE_BITE_ATTACK:
+				case TIGER_STATE_POUNCE_ATTACK:
+					if (!creature->Flags && item->TouchBits.Test(TigerBiteAttackJoints))
+					{
+						if	(item->Animation.AnimNumber == GetAnimNumber(*item, TIGER_ANIM_BITE_ATTACK) &&
```
Lwmte, line 213: I would add extra pair of brackets here before `||` operator, because eventually compiler may warn about logical operation precedence.

TombEngine/Game/collision/collide_room.cpp
```diff
@@ -847,19 +847,19 @@ void AlignEntityToSurface(ItemInfo* item, const Vector2& ellipse, float alpha, f
 
 	// Calculate extra rotation required.
 	auto extraRot = EulerAngles(
-		FROM_RAD(atan2f(forwardHeightDif, ellipse.y * 2)),
```
Lwmte, line 850: Any reason why `atan2f` was used here before?

Sezzary: Typo.

Sezzary (approved):

PR #984, opened 2023-01-31
Instancing - further renderer fixes

Kubsy:
- Instance static objects
- Greatly improved FPS
- Fix alpha sprites and refactor sprites.


TombEngine/Math/Geometry.cpp
```diff
@@ -260,4 +260,9 @@ namespace TEN::Math::Geometry
 
 		return false;
 	}
+
+	const int FastVectorsDistance(int x1, int y1, int z1, int x2, int y2, int z2)
```
Sezzary, line 264: Why not pass the actual vectors?

Sezzary: I also don't see it used anywhere.

MontyTRC89: It was added for saving costly SIMD load register operations. Yes, never used, I could remove it.

PR #987, opened 2023-02-01
Switches fix

Adngel:
I remade the branch.

This branch was doing:
- Fixes the error that unable the Valve switch (OCB 5) to untrigger objects.
- Removes the old code that blocked the Valve and Wall hole (OCB 5 and 6) to work with timed triggers.
- Polishes the distance position to use the switches.

It's required that builder fixes the assets for they work correctly. The next switches need to swap their ID states in WadTool.
- Big button switch (ocb 3)
- Valve switch (ocb 5)
- Wall Hole switch (ocb 6)

In this iteration, there are not editions to "SetAnimation ()" nor to the flipeffects.
The swt_Custom (OCB 7) is not summoning any particular animation, they just order to use the ItemFlags values (filled via Lua).
ItemFlags [4] on Animation
ItemFlags [5] off Animation
ItemFlags [6] distance to positionate

Remember you can set flags in the switch trigger.
- OnShot - Lara only will can use the switch once (OnAnimation)
- Timer - After Lara activates it, she won't be able to deactivate it. The switch will deactivate itself alone once the timer has run out.

OCBs:
0. LEVER SWITCH
1. SMALL LEVER SWITCH
2. SMALL BUTTON
3. BIG BUTTON
4. GIANT BUTTON
5. VALVE
6. WALL HOLE
7. CUSTOM (lua)
Anything else = Animation index of Lara "Switching on", and OCB+1 will be used for "Switching off".


TombEngine/Game/control/trigger.cpp
```diff
@@ -132,17 +132,14 @@ int SwitchTrigger(short itemNumber, short timer)
 				item->Timer = FPS * timer;
 			return 1;
 		}
-		if (item->TriggerFlags != 6 || item->Animation.ActiveState)
+		if (item->TriggerFlags >= 0 || item->Animation.ActiveState)
```
Sezzary, line 135: is `item->TriggerFlags` ever less than 0?

TombEngine/Objects/Generic/Switches/generic_switch.cpp
```diff
@@ -63,85 +76,115 @@ namespace TEN::Entities::Switches
...
 				return;
 
 			SwitchBounds.BoundingBox.X1 = bounds.X1 - 256;
 			SwitchBounds.BoundingBox.X2 = bounds.X2 + 256;
 
-			if (switchItem->TriggerFlags)
+			SwitchBounds.BoundingBox.Z1 = bounds.Z1 - 512;
+			SwitchBounds.BoundingBox.Z2 = bounds.Z2 + 512;
+
+			switch (switchItem->TriggerFlags)
```
Sezzary, line 88: Maybe it will be possible to define an `ObjectsCollisionBounds` object for each switch?

TombEngine/Game/control/trigger.cpp
Adngel: This is my concern.

TriggerFlags shall never be less than 0. But, if I remove that condition, then it only remains if (ActveState), in such situation, the code doesn't always enter and in the practice, the switches can't switched off.

item->TriggerFlags != 6 was meaning that the thread always enters except when ocb is 6 that in TEN is for the wall hole. I had to remove the wall hole restriction, but I couldn't remove the condition totally, so I put a value that won't cause problem, a negative one.

TombEngine/Objects/Generic/Switches/generic_switch.cpp
```diff
@@ -63,85 +76,115 @@ namespace TEN::Entities::Switches
...
+					break;
 			}
 
 			if (TestLaraPosition(SwitchBounds, switchItem, laraItem))
 			{
 				if (MoveLaraPosition(SwitchPos, switchItem, laraItem))
 				{
-					auto onAnim = LaraAnim::LA_WALLSWITCH_DOWN;
-					auto offAnim = LaraAnim::LA_WALLSWITCH_UP;
+					short onAnim = LaraAnim::LA_WALLSWITCH_DOWN;
```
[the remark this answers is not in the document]
Adngel: done

Adngel:
Also edited to include the function IsHeld(In::Action) and reorganize included by alphabetical order.

TombEngine/Objects/Generic/Switches/generic_switch.cpp
Adngel: I can try filled them also in the next case, like the SwitchPos.Z

Although I'm not sure which values, most of them could be similar (except for the big button and giant button that yes require a bigger bound like 512 one).

I'll update them and left the rest to 256 to see how it feels.

Adngel: At the end I made some more changes:
-  Some near of the wall are at BLOCK(0.2)
- Others that have more volume (like valve or small lever) I let them at BLOCK(0.25)
- The big button and giant button uses BLOCK(0.5) because Lara must get far from the wall to do the animation without digging in the object.
- For the custom at the moment I let the BLOCK(0.5), I don't know which animation will the builder use, so I let the big area, still in the future, if that happens to be insufficient, I could use the ItemFlags[7] for that, but I would prefer not to.

Sezzary (approved):

PR #979, opened 2023-01-30
Fix and improve TR3 tony

TokyoSU:
- Added BOSS effect.
- Removed global structure and variable BossData.
- Simplified the code and removed item->GetFlag, Item->SetFlag and item->TestFlag function which will be removed later to avoid confusion.
Don't merge this PR, it will be redone in another branch to avoid sophia commit.

PR #997, opened 2023-02-07
Fix tony not using boss effects.

TokyoSU:
- Fixed some comment of @Lwmte from PR #979 
- Require change from GetRandomControl() to Random::GenerateInt() @Sezzary
- Added some description and renamed TonyFlameType enum values.
- Removed GetFlagField and SetFlagField from Tony code, they will be removed later to avoid confusion.
- Added boss effects which tony was lacking.


TombEngine/Objects/TR3/Entity/TonyBoss.cpp
```diff
@@ -25,55 +25,70 @@ namespace TEN::Entities::Creatures::TR3
...
 	{
+		TONY_ANIM_WAIT, // Wait lara.
+		TONY_ANIM_RISE,
+		TONY_ANIM_FLY, // Real idle state.
+		TONY_ANIM_SHOOT_FIRE_RIGHT_HAND,
+		TONY_ANIM_SHOOT_CEILING, // with two hand.
+		TONY_ANIM_FLIPMAP, // Causes an shockwave explosion.
+		TONY_ANIM_DEATH
+	};
+
```
Nemoel-Tomo, line 70: 70 and 71 double blank line

TokyoSU: Fixed by @Sezzary :)

Adngel:
I had not any issue with Tony effects using this branch.

However one of the Tony attacks keep failing to hit Lara. (There straight fireball, is shoot at very high height so even if Lara is at same floor height than Tony, the firewall flies over here without touch Lara).


Aditionally, would be possible that Tony used the fire flame sprites for his fire effects? (Don't know very well how that work so I don't know if it's too difficult change or not).

TokyoSU:
> I had not any issue with Tony effects using this branch.
> 
> However one of the Tony attacks keep failing to hit Lara. (The straight fireball, is shoot at very high height so even if Lara is at same floor height than Tony, the firewall flies over her without touching Lara).

Yes i have this issue too. i will fix it :)

Kubsy:
I see the whole tony cpp file delete and replaced with a new. Tony still works right?

Kubsy (approved):

TokyoSU:
> I see the whole tony cpp file delete and replaced with a new. Tony still works right?

Yes, the new name is the coding convention which is applied.

Adngel (approved):

PR #999, opened 2023-02-09
Fixed shiva taking damage when in stone

TokyoSU:
- GetTargetOnLOS() was not using obj->HitRoutine when shooting enemies which was not triggered but drawed.


TombEngine/Game/control/los.cpp
```diff
@@ -183,32 +183,16 @@ bool GetTargetOnLOS(GameVector* origin, GameVector* target, bool drawTarget, boo
...
 							{
 								if (item->ObjectNumber >= ID_SMASH_OBJECT1 && item->ObjectNumber <= ID_SMASH_OBJECT8)
+								{
 									SmashObject(itemNumber);
+								}
 								else
 								{
-									switch (object->hitEffect)
-									{
-									case HitEffect::Blood:
```
Kubsy, line 191: Why was the case switch for effects removed? Is it not needed?

TombEngine/Game/control/los.cpp
```diff
@@ -183,32 +183,16 @@ bool GetTargetOnLOS(GameVector* origin, GameVector* target, bool drawTarget, boo
 							else
 							{
 								if (item->ObjectNumber >= ID_SMASH_OBJECT1 && item->ObjectNumber <= ID_SMASH_OBJECT8)
+								{
 									SmashObject(itemNumber);
+								}
 								else
 								{
-									switch (object->hitEffect)
```
Kubsy, line 189: Why was the case switch for effects removed? Is it not needed?

Sezzary (approved):

TombEngine/Game/control/los.cpp
TokyoSU: DefaultItemHit do that itself :)

Adngel:
This is not an issue of this branch, but I wonder, is possible use it to fix two things?

- The shiva statue is not moving down 1 click steps. (So I trigger it on a 1 click "pedestal" but it gets like locked in that sector).
- The shiva statue is not getting any damage from explossions.

These issues also happened in develop branch.

TokyoSU:
> This is not an issue of this branch, but I wonder, is possible use it to fix two things?
> 
> * The shiva statue is not moving down 1 click steps. (So I trigger it on a 1 click "pedestal" but it gets like locked, walking in rounds in that sector).
> * The shiva statue is not getting any damage from explossions.
> 
> These issues also happened in develop branch.

Fixed :)

Adngel (approved):

Kubsy (approved):

PR #1005, opened 2023-02-12
Electric Cleaner with vectors adngel

Adngel:
This is an updated version of Troye's electric cleaner.

Although it started from that branch, I've made an intensive re-work of the control function so I prefered present a different branch instead of push commits in the original one.

In this version:
- The object math are done by vectors instead of cases to move +x, -x, +y, -y so there are less duplicated rutines.
- There are more cases checks (check for box and non walkable flags, check for diagonal blocks, check for low ceiling gaps, and checks for sliding slopes). So builders will have more freedom to use this object.
- This version allows the cleaner move through slopes of 1 and 2 clicks, and uses Sezz's function to align the object to the floor.
- I've added more bit flags to the ItemFlags[1] with intention to edit them with Lua module.

ItemFlags[1].b0 = flagDoDetection (used by the object system)
ItemFlags[1].b1 = flagTurnRight (used by the object system)
ItemFlags[1].b2 = flagPriorityForward (for lua, makes the object try to continue going forward till it finds an obstacle).
ItemFlags[1].b3 = flagAntiClockWiseOrder (for lua, make the object choose betwee the left or right path as first option).
ItemFlags[1].b4 = flagStopAfterKill (for lua, make the object stops (OCB 0), or keep moving (OCB > 0), after kill Lara).

Notes:
- Please I got a bit confused with angles, this version works, but if you know more efficient functions please let me know.
- I try to test many varied cases, but if you find another one that escaped to me, please let me know with a pic to can replicate it and work on it.


TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,380 @@
...
+		
+		//Is a stopper tile? (There is still a shatter object).
+		if (col.Block->Stopper)
+			return false;
+
+		//If nothing of that happened, then it must be a valid sector.
+		return true;
+	}
+
+	Vector3 ElectricCleanerSearchDirections(ItemInfo& item, const Vector3& Dir1, const Vector3& Dir2, const Vector3& Dir3)
```
Kubsy, line 260: could it be possible to change variable names from Dir1, Dir2 ... to something meaningful? Like DirX, DirY etc

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,380 @@
...
+			break;
+
+		case ANGLE(-90.0f):
+			item.Pose.Position.x -= BLOCK(0.5f);
+			break;
+		}
+
+		if (GetCollidedObjects(&item, CLICK(1), true, CollidedItems, CollidedMeshes, true))
+		{
+			int lp = 0;
```
Kubsy, line 316: what is lp? :D

Adngel: Good question. I haven't got any idea. XD

That's a function I did not touch, however looks like this **lp** is just acting as an index, (same as if I call it **i** or **currentItem** in the logic I saw nothing wrong, although I don't know why the choice of word **lp**

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: No, DirX DirY would be more confusing:

The function is taking 3 directions, and they are called Dir1, Dir2 Dir3 because are the order.

1st it checks Dir1, if it's ok, returns it, if not, checks Dir2... etc...

And those Dirs, may be pointing north, south, east or west... the purpose of this SearchDirection is to can reuse the same functions but for different directions. So, any other name can not be linked to a cardinal direction, becaues that may change.

(You can see how they are called around line 124, each case order different directions, so I let Dir1,2,3 because that's something all have in common).

Adngel:
Lua module added:
https://github.com/TombEngine/Resources/tree/main/Lua%20Modules/Adngel

TombEngine/Game/collision/collide_room.h
```diff
@@ -126,6 +126,7 @@ CollisionResult GetCollision(ItemInfo* item);
 CollisionResult GetCollision(ItemInfo* item, short headingAngle, float forward, float down = 0.0f, float right = 0.0f);
 CollisionResult GetCollision(Vector3i pos, int roomNumber, short headingAngle, float forward, float down = 0.0f, float right = 0.0f);
 CollisionResult GetCollision(int x, int y, int z, short roomNumber);
+CollisionResult GetCollision(GameVector point);
```
Sezzary, line 129: Just an organisational thing, please make it `CollisionResult GetCollision(const GameVector& point);`

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
+#include "framework.h"
+#include "Objects/TR3/Trap/ElectricCleaner.h"
+
+#include "Game/collision/collide_item.h"
+#include "Game/control/box.h"
+//#include "Game/effects/debris.h"
```
Sezzary, line 6: If these includes are unneeded, they can be removed.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+		bool flagDoDetection		= ((item.ItemFlags[1] & (1 << 0)) != 0);
+		bool flagTurnRight			= ((item.ItemFlags[1] & (1 << 1)) != 0);
+		bool flagPriorityForward	= ((item.ItemFlags[1] & (1 << 2)) != 0);
+		bool flagAntiClockWiseOrder	= ((item.ItemFlags[1] & (1 << 3)) != 0);
+
+		auto col = GetCollision(item.Pose.Position.x, item.Pose.Position.y, item.Pose.Position.z, item.RoomNumber);
+
+		float yaw = TO_RAD(item.Pose.Orientation.y);
+
+		Vector3 ForwardDirection = Vector3(sin(yaw), 0, cos(yaw));
```
Sezzary, line 100: As a general rule, remember to use `auto` when instantiating class/struct variables, and the actual type name for variables representing primitives.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+		ForwardDirection.Normalize();
```
Sezzary, line 101: Also make these `forwardDirection` and `rightDirection`, since they are local variables.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+		
+		if (angleDifference > TO_RAD(rotationVel))
+		{
+			if (flagTurnRight)
+				item.Pose.Orientation.y -= rotationVel;
+			else
+				item.Pose.Orientation.y += rotationVel;
+
+			// Recalculate new difference to check if we should force align with axis for safety check.
+			angleDifference = abs (TO_RAD(goalAngle) - TO_RAD(item.Pose.Orientation.y));
```
Sezzary, line 114: No need for space after `abs`.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+			ItemNewRoom(itemNumber, probedRoomNumber);
+
+		auto radius = Vector2(object.radius, object.radius);
+		AlignEntityToSurface(&item, radius);
+
+		SpawnElectricCleanerSparks(item);
+		ElectricCleanerToItemCollision(item);
+	}
+
+	bool IsNextSectorValid(ItemInfo& item, const Vector3& Dir)
```
Sezzary, line 194: `dir` should be lowercase.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+			if (col.FloorTilt.y > 0 && col.FloorTilt.y > abs(col.FloorTilt.x))
+				slopeAngle = ANGLE(180.0f);
+			else if (col.FloorTilt.y < 0 && -col.FloorTilt.y > abs(col.FloorTilt.x))
+				slopeAngle = ANGLE(0.0f);
+
+			auto angleDir = FROM_RAD(atan2(Dir.x, Dir.z));
+			auto alignment = slopeAngle - angleDir;
+
+			//Is slope not aligned with the direction?
+			if ((alignment != 32768) && (alignment != 0) && (alignment != -32768))
```
Sezzary, line 239: Is it possible to demagic these angle values or are they like to avoid bugs?

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+				return false;
+		}
+
+		//Is diagonal floor?
+		if (col.Position.DiagonalStep)
+			return false;
+
+		//Is ceiling (square or diagonal) high enough?
+		int distanceToCeiling = abs(col.Position.Ceiling - col.Position.Floor);	
+		int cleanerHeight = 1024; //TODO change it for the collision bounding box height.
```
Sezzary, line 249: Use `BLOCK(1)`.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+		//Is there a pushable block?
+		if (CheckPushableList(MyPushablesList, detectionPoint.ToVector3()))
+			return false;
+
+		//If nothing of that happened, then it must be a valid sector.
+		return true;
+	}
+
+	Vector3 ElectricCleanerSearchDirections(ItemInfo& item, const Vector3& Dir1, const Vector3& Dir2, const Vector3& Dir3)
```
Sezzary, line 273: Use lowercase for these argument names too.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
```diff
@@ -0,0 +1,426 @@
...
+	{
+		auto pushableDistance = INFINITE;
+		for (int index = 0; index < PushablesList.size(); index++)
+		{
+			ItemInfo* currentObj = PushablesList[index];
+			
+			if (currentObj == nullptr)
+				continue;
+
+			Vector3 PushablePos = currentObj->Pose.Position.ToVector3();
```
Sezzary, line 417: `auto pushablePos`

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: I am not sure about the conversion. I know that:
- 32768 (in decimal) 
- is the same that 8000 in hexadecimal
- and the same that Lara's angleº + 180º

0 is Lara's angle, and -32768 I had to add it so it can consider also the slopes in opossite direction.

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: Done

TombEngine/Objects/TR3/Trap/ElectricCleaner.cpp
Adngel: So I put auto type for yaw and forwardDirection:

```c++
auto col = GetCollision(item.Pose.Position.x, item.Pose.Position.y, item.Pose.Position.z, item.RoomNumber);

auto yaw = TO_RAD(item.Pose.Orientation.y);

auto forwardDirection = Vector3(sin(yaw), 0, cos(yaw));
forwardDirection.Normalize();

auto rightDirection = Vector3(cos(yaw), 0, -sin(yaw));
rightDirection.Normalize();
```

Sezzary: Yes, and `yaw` can keep the `float`. :)

Kubsy (approved):

PR #1007, opened 2023-02-12
Addition of ID_MINECART_SWITCH

Stranger1992:
Now , rather than the switch for the minecart track being hardcoded to ANIMATING2 it has its own dedicated slot.

The switch is working in this file below.

[Minecart_Switch_Test.zip](https://github.com/MontyTRC89/TombEngine/files/10717586/Minecart_Switch_Test.zip)


Adngel (changes requested):
Yes, most of things looks like to be fine.

I just notice there is one line missing:

In code [TombEngine/Scripting/Internal/TEN/Objects/ObjectIDs.h](https://github.com/MontyTRC89/TombEngine/pull/1007/files#diff-7042d98392078136f2c045a398823097913140e172ef101869b89c1c84115da4)

The first half there is a long list of the objects commented, these comments in the lua folders, are not used by the TEN engine, but yes are used to generate the lua documentation and data.

So you should add your new ID slot also there. (I think that around line 632 will be ok, after the CROWDOVE_SWITCH line).


Over the line 632, the f

Stranger1992:
> Yes, most of things looks like to be fine.
> 
> I just notice there is one line missing:
> 
> In code [TombEngine/Scripting/Internal/TEN/Objects/ObjectIDs.h](https://github.com/MontyTRC89/TombEngine/pull/1007/files#diff-7042d98392078136f2c045a398823097913140e172ef101869b89c1c84115da4)
> 
> The first half there is a long list of the objects commented, these comments in the lua folders, are not used by the TEN engine, but yes are used to generate the lua documentation and data.
> 
> So you should add your new ID slot also there. (I think that around line 632 will be ok, after the CROWDOVE_SWITCH line).

Thanks! Done and pushed.

Adngel (approved):

Sezzary (approved):

Kubsy (approved):

PR #1013, opened 2023-02-19
Fix tr2 dragon

TokyoSU:
- WARNING: Fire breath is missing or not adapted to the dragon head which miss lara if she is not in front of the dragon!


TombEngine/Objects/TR2/Entity/tr2_dragon.cpp
```diff
@@ -234,6 +236,54 @@ namespace TEN::Entities::Creatures::TR2
 		ItemPushItem(item, laraItem, coll, 1, 0);
 	}
 
+
+	static void TriggerFireBreath(ItemInfo* item, const BiteInfo& bite, const Vector3i& speed, ItemInfo* enemy)
```
TokyoSU, line 240: This need to be changed to include the dragon joint rotation and allow him to attack lara on the left and right.

TombEngine/Objects/TR2/Entity/tr2_dragon.cpp
```diff
@@ -27,6 +29,7 @@ namespace TEN::Entities::Creatures::TR2
 	const auto DragonMouthBite = BiteInfo(Vector3(35.0f, 171.0f, 1168.0f), 12);
 	const auto DragonSwipeAttackJointsLeft  = std::vector<unsigned int>{ 24, 25, 26, 27, 28, 29, 30 };
 	const auto DragonSwipeAttackJointsRight = std::vector<unsigned int>{ 1, 2, 3, 4, 5, 6, 7 };
+	const auto DragonBackSpineJoints = std::vector<unsigned int>{ 21, 22, 23 };
```
Adngel, line 33: Meshes 21, 22 and 23 are the Dragon lower jaw and tongue. Maybe this was meant to be called DragonLowerJawJoints?

Adngel (commented):
I have to study the control function more deeply because I'm not understanding anything. @_@

However, I tested the dragon in game and I got a crash, when the dragon is defeated and Lara picks the dagger from its stomach. (Also her hand turns into a Lara's mesh 0 because I don't know yet what swap mesh use to Lara's hand with the dagger). 

When I get more info I'll come back to you.

TombEngine/Objects/TR2/Entity/tr2_dragon.cpp
TokyoSU: No it's the spine joints in the dragon back, OG remove it from the code to avoid it being seen in-game

TokyoSU:
> However, I tested the dragon in game and I got a crash, when the dragon is defeated and Lara picks the dagger from its stomach. (Also her hand turns into a Lara's mesh 0 because I don't know yet what swap mesh use to Lara's hand with the dagger).

Because i've not checked this part :x

Adngel:
I re-wrote the routine for the dagger pickup, I think now is much more easy to pick the dagger as it uses the automatic alignment function and influence areas, so no more spam ACTION key  while rotating to activate the dagger pickup. (editado)

I think there are some issues with the collision (the creature is big and basically when the Dragon collides with Lara, it teleports her around more than push her.

I'll try to look at that tomorrow, but apart of that, I think is getting close to be completed

- The crash for the collision function in the back part, is fixed.
- The problem with the visible back part is gone.
- The warning message in the log has been solved too.
- The dagger pickup routine now works better.
- I added some shockwave effect to its attack so Lara now can't stay safe in a dragon blind spot.
- I added the ocb separation (OCB 0 to can directly kill the dragon, OCB 1 to make that Lara needs pick the dagger to can kill it).
- and I changed the Random generators to use the normalized values

Adngel:
I`ve though that for the collision, if the ItemPushItem keep giving issues, I could hack the test collision during the defeat state. To only test the main spheres along the dragon spine.  Ignoring the ones of the legs, maybe that way it could reduce the collision pushes during the dagger pickup. 

Could be another solution?

Adngel:
> I`ve though that for the collision, if the ItemPushItem keep giving issues, I could hack the test collision during the defeat state. To only test the main spheres along the dragon spine. Ignoring the ones of the legs, maybe that way it could reduce the collision pushes during the dagger pickup.
> 
> Could be another solution?

I saw the Sezz's branch about the soft collision, and I think it worth wait for that implemention instead of bringing more hacks to the dragon collision function.

Adngel:
Corrupted PR, I'm re-making on a fresh PR.

PR #1017, opened 2023-02-23
Fix direction generation

Sezzary:
- Fix random direction generation function having a bias toward cube corners.
- Fix point in sphere generation function having a bias toward positions closer to the centre of the sphere.
- Add `GeneratePoint2DInSquare()`.
- Add `GeneratePoint2DInCircle()`.
- Add `GeneratePointOnSphere()`.

### DIFF
--- a/TombEngine/Math/Random.cpp
+++ b/TombEngine/Math/Random.cpp
@@ -48,27 +48,38 @@ namespace TEN::Math::Random
 		float x = GenerateFloat(-semiangleInDeg, semiangleInDeg) * RADIAN;
 		float y = GenerateFloat(-semiangleInDeg, semiangleInDeg) * RADIAN;
 		float z = GenerateFloat(-semiangleInDeg, semiangleInDeg) * RADIAN;
-		auto matrix = Matrix::CreateRotationX(x) * Matrix::CreateRotationY(y) * Matrix::CreateRotationZ(z);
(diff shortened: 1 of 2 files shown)


TombEngine/Math/Random.cpp
```diff
@@ -26,54 +26,112 @@ namespace TEN::Math::Random
 
 	Vector2 GenerateDirection2D()
 	{
-		auto vector = Vector2(GenerateFloat(-1.0f, 1.0f), GenerateFloat(-1.0f, 1.0f));
-		vector.Normalize();
-		return vector;
+		float angle = GenerateFloat(0.0f, PI_MUL_2); // Generate angle in full circle.
```
Lwmte, line 29: Wouldn't this approach be slower than just generating from `(-1, 1)` range directly? Although if this function will be rarely used, it shouldn't be a big problem.

Lwmte (commented):
I see no problems with this pull request, apart from potential perfo problems which I mentioned above, but this is a minor issue and can be dealt with if that function becomes real bottleneck. Unless it doesn't happen, it can be left as is.

TombEngine/Math/Random.cpp
Sezzary: Using the old `(-1, 1)` range would result in a bias toward directions closer to square corners, which is what I wanted to avoid in the 3D variant of this function.

Kubsy (approved):

PR #1019, opened 2023-02-26
slamming door backport

Kubsy:
works but we need to figure out:

- why blood is not appearing - **done** 
- if we should name it TeethSpikesDoor or just SlammingDoors 
- Add the slamming door ID to the internal/ObjectIDs.h for Lua usage - **done**
- Maybe reduce the duplication by having the array of the bites - **this didn't work.**


TombEngine/Objects/TR1/tr1_objects.cpp
```diff
@@ -22,6 +22,7 @@
 
 // Traps
 #include "Objects/TR1/Trap/DamoclesSword.h"
+#include "Objects/TR1/Trap/TeethSpikeDoor.h"
```
Adngel: If the slot name has been changed, does it need change the file and functions names too? To SlammingDoor

TombEngine/Objects/TR1/tr1_objects.cpp
```diff
@@ -188,6 +189,15 @@ static void StartTrap(ObjectInfo* obj)
...
 		obj->SetupHitEffect(true);
 	}
+
+	obj = &Objects[ID_TEETH_SPIKE_DOOR];
+	if (obj->loaded)
+	{
+		obj->initialise = InitialiseTeethSpikeDoor;
+		obj->control = ControlTeethSpikeDoor;
+		obj->collision = GenericSphereBoxCollision;
+		obj->SetupHitEffect(true);
```
Adngel, line 199: Remember me to check for this. This genericSphereBoxCollision uses the ItemFlags[3] to read Damage and apply it to Lara, so right now, this system is activating doDamage 2 times. Once with this collision function that applies 0 damage and other in control function that applies the 400.

There may be other functions more suitable for collisions in the other traps  need to check. Or it could adapt the damage in the itemFlags to avoid the duplicated doDamage call.

Adngel (commented):
I tested yesterday with TR1 Midas trap and it worked, when I were at home I'll test with the other tr3 slamming doors.

TombEngine/Objects/TR1/tr1_objects.cpp
Kubsy: hmm there are no other functions for trap collision as far as I saw.

TombEngine/Objects/TR1/Trap/SlammingDoors.cpp
```diff
@@ -0,0 +1,68 @@
+#include "framework.h"
+#include "Objects/TR1/Trap/SlammingDoors.h"
+
+#include "Game/control/Box.h"
+#include "Game/effects/effects.h"
+#include "Game/Lara/lara.h"
+#include "Math/Objects/Vector3i.h"
```
Sezzary, line 7: Math can be used like a library. You only need to import `Math/Math.h`.

TombEngine/Objects/TR1/Trap/SlammingDoors.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "Math/Random.h"
+
+using namespace TEN::Math;
+
+namespace TEN::Entities::Traps::TR1
+{
+	constexpr auto SLAMMING_DOORS_DAMAGE = 400;
+	constexpr auto LARA_RADIUS = 100;
```
Sezzary, line 15: This constant already exists in `lara.h`.

TombEngine/Objects/TR1/Trap/SlammingDoors.cpp
```diff
@@ -0,0 +1,68 @@
...
+			if (item.Animation.TargetState != SLAMMINGDOORS_DISABLED)
+			{
+				item.Animation.TargetState = SLAMMINGDOORS_DISABLED;
+				item.ItemFlags[0] = 0;
+			}
+		}
+
+		AnimateItem(&item);
+	}
+}
```
Sezzary, line 68: Remember to have a newline at the end of the file, otherwise github complains.

TombEngine/Objects/TR1/Trap/SlammingDoors.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+struct ItemInfo;
+struct BiteInfo;
+
+namespace TEN::Entities::Traps::TR1
+{
+	void InitialiseSlammingDoors(short itemNumber);
+	void ControlSlammingDoors(short itemNumber);
+}
```
Sezzary, line 10: Add newline at the end here as well.

TombEngine/Objects/TR1/tr1_objects.cpp
Adngel: Resolved on Discord chat.

Adngel (approved):
I tested and reviewed, I've got no more objections with this branch, green light from me.

TombEngine/Objects/TR1/Trap/SlammingDoors.cpp
Adngel: And LARA_RADIUS is unused in this code now, because the blood spawning is sorted by the trap collision function.

Sezzary (approved):

PR #1020, opened 2023-02-26
Stranger1992/bass engine upgrade

Stranger1992:
Updated UnReal bass engine to 2.4.17 ( https://www.un4seen.com/ ) 

[bass updated dll.zip](https://github.com/MontyTRC89/TombEngine/files/10834159/bass.updated.dll.zip)

Also added in the ability for TEN to use sound samples at 44.1khz instead of 22.5khz if they are present.


TombEngine/Sound/sound.cpp
```diff
@@ -90,10 +102,12 @@ bool LoadSample(char* pointer, int compSize, int uncompSize, int index)
 	BASS_SampleGetInfo(sample, &info);
 	int finalLength = info.length + 44;	// uncompSize is invalid after 16->32 bit conversion
 
-	if (info.freq != 22050 || info.chans != 1)
+
 	{
-		TENLog("Wrong sample parameters, must be 22050 Hz Mono", LogLevel::Error);
-		return false;
```
Kubsy, line 96: why was `return false` deleted?

the return false makes sure it will exit the function if it couldn't load samples with either frequency (I haven't looked at the function but by the if statement and function name I can guess :p).

Stranger1992: done.

TombEngine/Sound/sound.cpp
```diff
@@ -90,9 +102,12 @@ bool LoadSample(char* pointer, int compSize, int uncompSize, int index)
...
 
-	if (info.freq != 22050 || info.chans != 1)
+
 	{
-		TENLog("Wrong sample parameters, must be 22050 Hz Mono", LogLevel::Error);
+		if (info.freq != TRLE_QUALITY_SAMPLES && info.freq != HIGH_QUALITY_SAMPLES || info.chans != MONO)
+		{
+			TENLog("Incorrect sample parameters, must be either 22050 Hz or 44100 Hz (Mono)", LogLevel::Error);
+		}
 		return false;
```
Kubsy, line 111: You need to put return false inside if statement

TombEngine/Sound/sound.cpp
```diff
@@ -90,9 +90,12 @@ bool LoadSample(char* pointer, int compSize, int uncompSize, int index)
...
```
Kubsy, line 99: I think this return false needs to be inside if statement.

Lwmte: Moreover, why additional bracket block was introduced without any purpose? This whole code part looks weird to me now.

TombEngine/Sound/sound.h
```diff
@@ -38,6 +38,18 @@ enum class SoundTrackType
 	Count
 };
 
+enum SampleRate
```
Lwmte, line 41: This enum does not make sense. TEN internally resamples a sound according to a pitch value in sound info. If you allow 2 separate sample rates, a major sound engine refactor must be performed so that engine knows original sample rate for every loaded sample. We won't dig into that now. Remove this.

TombEngine/Sound/sound.h
```diff
@@ -38,6 +38,18 @@ enum class SoundTrackType
...
+{
+	TRLE_QUALITY_SAMPLES = 22050,
+	HIGH_QUALITY_SAMPLES = 44100,
+};
+
+enum Channels
```
Lwmte, line 47: Sound samples which are playing in a 3D space, which is essentially all the sound engine does, cannot be stereo. Therefore, this enum makes no sense as well.

Stranger1992:
I have removed all commits with my controversially incorrect code. The commits that remain are just the updated header files

PR #1024, opened 2023-02-28
Stranger1992/tr5 imp refactor

Stranger1992:
**Fix ID_IMP :**

- Death is now triggered properly. 
- Imp throws ID_IMP_ROCK at Lara
- Imp is now scared of fire.

**OCB:**

- CLIMB_UP_START = 1,
- BARREL_ROLL_START = 2,
- THROW_STONES = 3

Imp can be found here : https://github.com/TombEngine/Resources/blob/main/Wad2%20Objects/tr5_Imp.wad2


TombEngine/Objects/TR5/Entity/tr5_imp.cpp
```diff
@@ -96,206 +135,252 @@ namespace TEN::Entities::Creatures::TR5
...
+				creature->Mood = MoodType::Escape;
+			CreatureMood(item, &AI, true);
 
-				if (item->Animation.ActiveState == IMP_STATE_SCARED)
-					creature->Mood = MoodType::Escape;
+			angle = CreatureTurn(item, creature->MaxTurn);
 
-				CreatureMood(item, &AI, true);
+			if (Wibble & 0x10)
+				item->SetMeshSwapFlags(0x400);
```
Adngel, line 240: This 0x400 can be demagic

0x400 hex, in decimal is 1024, but in binary is: 0000 0100 0000 0000

The mesh swaps use those binary values to indicate which mesh must be swap. In this case the 1 is in the 10th bit (starting by 0, 1, 2...). And if you check the Wad Tool, you'll see the 10th mesh is the head.

So it can be created a constant value at the top giving a name and the mesh number, to later use that name instead of the 0x400 that may be harder to identify.

If you want to see examples, Guide code is full of these.

So it could change it with:
```c++
/Adding the initialization at top with the other constants
const auto ImpHeadSwapJoints	 = std::vector<unsigned int>{ 10 };  

//in the SetMeshSwapFlags
item->SetMeshSwapFlags( ImpHeadSwapJoints );
```

It's using a vector, because it can keep more than 1 mesh. Imagine if the binary number were:
0000 0100 0010 0100
Then the swap mesh would change the 2nd, 5th and 10th meshes (left leg, body and head).

I hope this info helps.

TombEngine/Objects/TR5/Entity/tr5_imp.cpp
```diff
@@ -96,206 +135,252 @@ namespace TEN::Entities::Creatures::TR5
...
+				{
+					item->Animation.TargetState = IMP_STATE_SCARED;
 					break;
+				}
 
-				case IMP_STATE_ATTACK_1:
-				case IMP_STATE_ATTACK_2:
-					creature->MaxTurn = -1;
+				if (AI.bite && AI.distance < IMP_IDLE_ATTACK_RANGE && item->TriggerFlags < 10)
+				{
```
Adngel, line 282: At the end of the condition, is checking that the item->TriggerFlags < 10

TriggerFlags is the OCB, I the planned ocbs for this creature were 0,1, 2 and 3.  So this last condition case shall not happens, right?

TombEngine/Objects/TR5/Entity/tr5_imp.cpp
```diff
@@ -96,206 +135,252 @@ namespace TEN::Entities::Creatures::TR5
...
+				else if(item->TriggerFlags == IMP_OCB_THROW_STONES)
+				{
+					item->Animation.TargetState = IMP_STATE_THROW_STONES;
+				}
+				else if (AI.distance > IMP_WALK_RANGE)
+				{
+					item->Animation.TargetState = IMP_STATE_RUN;
+				}
+				else if (AI.distance > IMP_LOW_ATTACK_RANGE || item->TriggerFlags < 10)
+				{
```
Adngel, line 301: This second part of the condition, would always trigger, (OCB is always going to be lower than 10) that would make the first part of the condition a bit unuseful.

Maybe you want delete that 2nd part.

TombEngine/Objects/TR5/Entity/tr5_imp.cpp
```diff
@@ -96,206 +135,252 @@ namespace TEN::Entities::Creatures::TR5
...
+				break;
+			case IMP_STATE_JUMP_ATTACK:
+				RotateTowardTarget(*item, AI, IMP_ATTACKJUMP_OR_STONE_TURN_RATE_MAX);
+
+				if (!(creature->Flags & 1) && item->TouchBits.Test(ImpRightHandBite.meshNum))
+				{
+					DoDamage(creature->Enemy, IMP_ATTACK_DAMAGE);
+					CreatureEffect(item, ImpRightHandBite, DoBloodSplat);
+					creature->Flags |= 1;
 				}
```
Adngel, line 343: "creature->Flags & 1" is just checking if the flags is "creature->Flags == 1"

and "creature->Flags |= 1" is the reseting the variable, the same as "creature->Flags = 0"

I know it could be done with a bool, but every entity (that do melee attacks) also use the Flags like an short, (someone even use "creature->Flags =2".

Not a big issue, I leave up to you if prefer changing it or leave it with bitwise operations.

Adngel (commented):
I reviewed the code.

I saw it working and compiles, although I pointed some points that could changed, but nothing crucial though.

Apart or that I've got no objection with this code. I'll do some ingame test before give it green light.

Adngel (approved):

Adngel (approved):

PR #1025, opened 2023-03-01
Tr3 boateffect

Nemoel-Tomo:
Currently the code just checks if the wake direction is left, right or center and then it adds its segments and previous segments to the associated wave direction.
Currently only one streamer of each direction can be active at the same time. One in the whole game, sadly not per object.
If there is more than one streamer with the same direction the segments get mixed up between those. 
For the vehicles it should be no problem as long as only one vehicle is active.
Maybe someone find a way to give every stream an ID or attach the itemID to the stream where all related segments and previous segments are allocated.

The streams work on TR2 speedboat, TR3 rubber boat, kajak, UPV. It also works on quadbike, snowmobile, minecart, jeep and motorcycle.  (1 click deep water). Every vehicle should be tested.

Speedboat with OCB 0  additionally makes oldschool TR2 foam effect, and with ocb1 it makes a TRNG effect

### DIFF
--- a/TombEngine/Game/Lara/lara.cpp
+++ b/TombEngine/Game/Lara/lara.cpp
@@ -450,11 +450,11 @@ void LaraControl(ItemInfo* item, CollisionInfo* coll)
 		};
 
 		// Spawn Commander Video rainbow.
-		StreamerEffect.GrowStreamer(item->Index, 0, streamerPositions[0], direction, orient2D, color0, width, life, scaleRate);
-		StreamerEffect.GrowStreamer(item->Index, 1, streamerPositions[1], direction, orient2D, color1, width, life, scaleRate);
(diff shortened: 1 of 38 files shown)


Nemoel-Tomo:
Please do not merge until it is clear what the other 93 commits are about!!!!!

Kubsy:
> Please do not merge until it is clear what the other 93 commits are about!!!!!



> Please do not merge until it is clear what the other 93 commits are about!!!!!

weird because the changed files show only what you have done, not about the other ones.

Adngel:
Shall we recreate the 500 lines of this line in a new branch? (tomorrow I've got time, can do a new one and copy the changes from this one if needed).

Nemoel-Tomo:
let's keep it like this for now. I need to rewrite some code lines and maybe Sezz find a way to make more streamers at the same time for multiple objects possible. When everything is done, it really should be transferred to a new branch

Sezzary (approved):

Kubsy (approved):

PR #1029, opened 2023-03-10
Adngel Pushables review

Adngel:
This branch rewrites the pushables codes:
- To fix bugs.
- To improve performance.
- To improve the code readability and reduce code complexity.
- To change and add some features.

**Fixed bugs:**
- Now it only activate/deactivate the stopper flag in the sector where it's and the sector where it's moving. It won't activate the stopper flag in the surounding tiles anymore. 
- Now it doesn't edit the blocked status of the floor grey flags. (If the block must really do blocking grey flags, a new function shall be done, as the previous one was in risk of infinite recursivity bug if there were several grey flags non connected in the same sector island where the pushable object was. Besides it also was covering sectors beyond the object).
- Now the pushable will update its room number when Lara moves it, and not just for the stacked pushables.
- Now if there is an object over the Climbable pushable, and Lara moves the lower pushable, the stacked one will move with it.

**Remaining bugs:**
- If there is a pushable stacked over another pushable, and Lara is also over another pushable, she won't be able to interact with the stacked pushable. (Bridge system error)
- In some cases the stacked pushables may loose their collision.  (Bridge system error)
- The pushable collision may get lost if mover over trapdoors. (Bridge system error)
- When using the TR1-TR3 animations, the pushable is displaced a bit at the beggining of push and at the of pull. The position later is sorted but visually there is this wird gap. (Surelly a inssue with the animation from the Wad).

**Changes and additions:**
- Segregation between Pushable_Objects and Pushable_Object_Climbable. The pushable_objects are statues and other props that Lara can move but she can't climb (they use object collision). The Pushable_Object_Climbabe are for blocks, tables, funitures, that Lara can move, climb and walk over them, (they use bridge collision system). With this change, it was deleted the ocbs related to the object climbable height as this is calculated automatically from the object bounding box.
- OCBs, the pushable objects now uses:
   - ocb +1 = The block can fall.
   - ocb +2 = Lara will positionate at the center of the object to interact with it.
   - ocb +4 = The object will float on water.
   - ocb +8 = The pushable will use Lara TR1-TR3 animation system (need animations in wad).
- Implementation of a sound map so they can produce different sounds depending of the floor material, (but deleted the personal sound copy that each pushable had, now all pushables in the level will use the same ones.
- Ability for the pushables to float in water need the buoyancy flag (ocb +4 ) for it to work. It reacts to flipmaps changes.

- New Debug page to check the floor data flags. (If there is a flag, it will spawn a sphere that is drawn when Lara is close, the color of the sphers depends of the flag).
- Change the box debug function to paint the isolated boxes (the ones created with grey flag) in red if they are Blocked and Green if they are UnBlocked.

- Updated the Electric Cleaner code to delete the pushable objects search method. (As the stopper flag is working correctly, Electrict cleaner already reads them so it can avoid hitting pushables in most of the cases).

**Pending**
- Fixing the stack system, but most of its bugs are require study and modifications in the floor_data code, but due to the already long lenght of this branch, and the potential plan to re-work the Bridge system, this issues fixes got postponed.
- Fix the wad TR1-TR3 pushable animations.
- Fill the sound map with the different sounds.
- Implement the lua functions to allow builders customize some of the pushables features.

I open the PR for feedback and tests.


Stranger1992:
@Adngel I can implement the new sound slots if needed.

Adngel:
> implement the new sound slot

Adngel:
> @Adngel I can implement the new sound slots if needed.

Is one of the missing things. The sound map is ready to hold three kind of sounds:

- Move: is the loop sound that happens while the pushable is being moved.
- Stop: is a single sound that happens when the pushable movement has finished.
- Fall: is a single sound that happens when the pushable has been pushed through a fall and has collided with the floor.

The new system also checks the floor material so it can run the sounds linked to that material. There are 22 materials, but without counting the customs, there are 14. One is for stone that is done with the default sounds. 
So it still misses 13 x 3 = 39 sounds.

They are applied in the function: void InitializePushablesSoundsMap() (line 726 of the tr5_pushableblock.cpp ) The code only needs the sound indices. Unlike other entities, the sounds for this object are uniquely applied by code, not animation, (so it may be necessary to imput the sounds manually in the Tomb Editor).

Stranger1992:
@Adngel sound implementation done.

[Adngel Pushable TE files.zip](https://github.com/MontyTRC89/TombEngine/files/10975569/Adngel.Pushable.TE.files.zip)
You will also need these two files. Drop ```trCatalog.xml``` into your Tomb Editor / Catalogs folder and use the other XML in Tomb Editor itself. PS: The volumes / pitch variance / random etc is not done as that will be next.

Do let me know if you need anything else :)

TombEngine/Game/items.cpp
```diff
@@ -838,61 +838,56 @@ void DefaultItemHit(ItemInfo& target, ItemInfo& source, std::optional<GameVector
...
-					 pos.x & maskLower8Bits | mask9Bit,
-					 pos.y,
-					 pos.z & maskLower8Bits | mask9Bit
-					);
+	// Return collision block center.
+	// TODO: No bitwise operations.
+	return Vector3i(
+		pos.x & BIT_MASK_LOWER_8 | BIT_MASK_9,
+		pos.y,
+		pos.z & BIT_MASK_LOWER_8 | BIT_MASK_9);
```
Adngel, line 853: It can be done with maths, but I let the bitwise operation because although ugly, they were more optimal.

So I created this function to move the bitwise ugliness here instead of the object control function.

If decided to apply with maths in a readable way, I think this function could be discarded (is not used a lot, I think like 4 or 5 times along the engine).

TombEngine/Game/items.cpp
```diff
@@ -838,61 +838,56 @@ void DefaultItemHit(ItemInfo& target, ItemInfo& source, std::optional<GameVector
...
-	{
-		return itemA.Pose.Position.z < itemB.Pose.Position.z;
-	}
+	if (item0.Pose.Position.x == item1.Pose.Position.x)
+		return (item0.Pose.Position.z < item1.Pose.Position.z);
 	else 
-	{
-		return itemA.Pose.Position.x < itemB.Pose.Position.x;
-	}
+		return (item0.Pose.Position.x < item1.Pose.Position.x);
```
Adngel, line 864: This goes against the project coding convention. Lwmte in the civvy branch, asked me to keep the brackets in the if-else and if-elseif structures.

https://github.com/MontyTRC89/TombEngine/issues/294  

> However, if you have only one line enclosed (i. e. only foo), and there is no following else or else if conditions, it is currently allowed (but not recommended) to omit brackets

Sezzary: If there are only two conditions, both contain single-line blocks, and the second is an open `else`, I've tended to skip using braces in such instances for some time. Maybe we should get further clarification?

TombEngine/Game/items.cpp
Sezzary: It's a fairly high-level function, so optimisation of that sort isn't too necessary.

Sezzary: The function is also handy so I wouldn't discard it. :)

Stranger1992:
- [ ] Lara will walk through walls and get stuck in the geometry if she pushes a `pushable_object_climbable` forward 1 sector then pulls it back :

image

Stranger1992:
- [ ] Lara goes too far forward when pushing an OCB 1 `PUSHABLE_OBJECT` from a height and will fall down with it


Uploading 2023-05-10_21-04-21.mp4…

Stranger1992:
The pushable sounds work really nicely. I need to push changes to the Tomb Editor catalog though but as a demo: see here

https://discord.com/channels/1072511394735464498/1072569001986773023/1105995650199474287

Adngel:
PR is right now deprecated, changes pending to be translated into a new clean branch that is updated to current engine status.

Adngel:
Closed for rework.

PR #1031, opened 2023-03-12
Corpse object

Nemoel-Tomo:
Fixed Compsognathus idle animation before dying.
A new object has been created in slot #1086. ID_CORPSE.
OCB 0 is a lying carcass like in TR3 "Crash Side" level in T-Rex hideout.
With OCB 1 it hangs in the air like the raptor carcass in TR3 level "Crash Side". If Lara shoots it, it will fall down.
The carcass effect can be triggered or anti-triggered and has a green smelly smoke visual effect.
The Compsognathus dinos will attack the activated corpse if Lara is unreachable.

ItemFlags[1] = activate object attribute:
		CadaverLying = 0,
		CadaverHanging = 1,
		CadaverFalling = 2,

I have removed the effectID::Corpse and moved the visuals to object ID_CORPSE only. Trigger/antitrigger the effect on the corpse object as your wish.


Nemoel-Tomo:
[raptor-cadaver and compy.zip](https://github.com/MontyTRC89/TombEngine/files/10951586/raptor-cadaver.and.compy.zip)

Kubsy (approved):

PR #1034, opened 2023-03-18
Simple animation refactors

Sezzary:
Simple refactors. Makes simplifications, switches to Quaternions or EulerAngles where appropriate, adds getters, etc.

Slowly want to move toward removing that giant AnimData vector and keep animation data for each object separately.


Sezzary:
> will test entities which borrow anims from other slots

Don't need to, I've decided to leave that for the next PR. In the end, the steps will be:
1. Simple refactors (this PR).
2. Add new `SetAnimation()` overload with proper handling of animations located in different slots.
3. Adopt `GetAnimData()` everywhere to eliminate inappropriate `g_Level.Anims` access.
4. Address inappropriate `g_Level.Frames` access in a similar way.
5. Eliminate `g_Level.Anims` and `g_Level.Frames` and store animation data for each object inside `ObjectInfo`.

Kubsy (approved):

PR #1036, opened 2023-03-20
Added the LUA SetFlags() function

davidmarr:
Added the LUA SetFlags() function. It allows you to change the flags options of a string

### DIFF
--- a/TombEngine/Scripting/Internal/TEN/Strings/DisplayString/DisplayString.cpp
+++ b/TombEngine/Scripting/Internal/TEN/Strings/DisplayString/DisplayString.cpp
@@ -132,7 +132,12 @@ void DisplayString::Register(sol::table & parent)
 		// @function DisplayString:GetPosition
 		// @treturn int x x-coordinate of the string
 		// @treturn int y y-coordinate of the string
-		ScriptReserved_GetPosition, &DisplayString::GetPos
+		ScriptReserved_GetPosition, &DisplayString::GetPos,
+
+		/// Set the display string's flags 
+		// @function DisplayString:SetFlags
+		// @tparam Table table the new table with display flags options
+		ScriptReserved_SetFlags, & DisplayString::SetFlags
 	);
 }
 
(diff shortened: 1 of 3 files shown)


TombEngine/Scripting/Internal/TEN/Strings/DisplayString/DisplayString.cpp
```diff
@@ -181,6 +186,16 @@ std::string DisplayString::GetKey() const
 	return s.m_key;
 }
 
+void DisplayString::SetFlags(sol::table const& flags) {
```
Sezzary, line 189: Please follow TEN conventions. ;) Braces should always be on separate lines, `const` should be placed before `sol::table&` (all of scripting had this wrong from the beginning).

TombEngine/Scripting/Internal/TEN/Strings/DisplayString/DisplayString.cpp
```diff
@@ -181,6 +186,16 @@ std::string DisplayString::GetKey() const
...
+	UserDisplayString& s = s_getItemCallback(m_id).value();
```
Sezzary, line 190: Please aim to use variable names longer than a single character as they get pretty vague.

davidmarr: I used the same convention as the other functions. They use a variable **s** to set a parameter or return a value. However, I can modify the function I created to be clearer

TombEngine/Scripting/Internal/TEN/Strings/DisplayString/DisplayString.cpp
```diff
@@ -132,7 +132,12 @@ void DisplayString::Register(sol::table & parent)
 		// @function DisplayString:GetPosition
 		// @treturn int x x-coordinate of the string
 		// @treturn int y y-coordinate of the string
-		ScriptReserved_GetPosition, &DisplayString::GetPos
+		ScriptReserved_GetPosition, &DisplayString::GetPos,
+
+		/// Set the display string's flags 
+		// @function DisplayString:SetFlags
+		// @tparam Table table the new table with display flags options
+		ScriptReserved_SetFlags, & DisplayString::SetFlags
```
hispidence, line 140: Please add an example just to make it clear that the `Strings.DisplayStringOption` flag should be used (assuming they are)

hispidence (approved):

Sezzary (approved):

PR #1037, opened 2023-03-21
IsSoundEffectplaying lua implementation

Kubsy:
Implemented the above command.


TombEngine/Sound/sound.cpp
```diff
@@ -628,6 +628,15 @@ int Sound_EffectIsPlaying(int effectID, Pose *position)
 	return -1;
 }
 
+bool Sound_EffectIsPlaying(int effectID)
+{
+	if (effectID == -1)
+		return;
```
Sezzary, line 634: This doesn't really serve any purpose. The function should return `false` if `channelIndex` is -1. ;)

TombEngine/Sound/sound.cpp
```diff
@@ -628,6 +628,15 @@ int Sound_EffectIsPlaying(int effectID, Pose *position)
 	return -1;
 }
 
+bool Sound_EffectIsPlaying(int effectID)
```
Sezzary, line 631: Can rename to `IsSoundEffectPlaying()` to comply with conventions. It doesn't need to be an overload of the legacy function since it does something different.

TombEngine/Sound/sound.cpp
Kubsy: done :)

TombEngine/Sound/sound.cpp
Kubsy: done :)

Sezzary (approved):

PR #1068, opened 2023-04-15
2-way puzzle holes

Nemoel-Tomo:
Put a switch trigger for the puzzle hole on the square instead of a key trigger to make it a 2-way puzzle hole. Else it works like in OG.
New Lara object in the comment below.

Fix spiky wall speed. Insert speed value into OCB. you can change the speed ingame by changing itemflag[0].


Nemoel-Tomo:
[Lara-remove-puzzle.zip](https://github.com/MontyTRC89/TombEngine/files/11239563/Lara-remove-puzzle.zip)

Sezzary (approved):

Kubsy (approved):

PR #1074, opened 2023-04-18
Sprites in 2D space

Sezzary:
Simple base for the functionality. Still needs Lua integration and organised handling which can be done later.


TombEngine/Renderer/Renderer11Draw2D.cpp
```diff
@@ -373,74 +375,147 @@ namespace TEN::Renderer
...
+			vertexPoint = TEN::Utils::Convert2DPositionToNDC(vertexPoint);
+		}
+
+		g_Renderer.PrintDebugMessage("%.3f, %.3f", vertexPoints[0].x, vertexPoints[0].y);
+
+		// Define renderer vertices.
+		auto vertices = std::array<RendererVertex, VERTEX_COUNT>{};
+
+		// Vertex 0
+		vertices[0].Position = Vector3(vertexPoints[0]);
```
Kubsy, line 416: this might be a weird idea but how about having all of the vertex 0-3 setup in a for loop? 

For example if the number is odd, then process for first x and second y and if it is even then process for second UV constraint

TombEngine/Renderer/RendererSprites.cpp
```diff
@@ -4,6 +4,11 @@
 
 namespace TEN::Renderer 
 {
+	void AddSpriteScreenSpace()
```
Kubsy, line 7: hmm add a TODO comment?

Sezzary: yes

Kubsy (approved):

PR #1087, opened 2023-04-25
Check levels in Gameflow.lua

davidmarr:
Check if there is the title and at least one playable level to avoid crashing the game

### DIFF
--- a/TombEngine/Game/control/control.cpp
+++ b/TombEngine/Game/control/control.cpp
@@ -271,6 +271,13 @@ unsigned CALLBACK GameMain(void *)
 	else
 		g_Renderer.RenderTitleImage();
 
+	//Check if there is the title in Gameflow.lua
+	TENLog("Number Levels: " + std::to_string(g_GameFlow->GetNumLevels()), LogLevel::Info);
+	if (g_GameFlow->GetNumLevels() == 0)
+	{
+		TENLog("No title found. Please check 'Gameflow.lua' file.", LogLevel::Error);
+		WinClose();
+	}
 
 	// Execute the Lua gameflow and play the game.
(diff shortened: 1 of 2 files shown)


Kubsy (approved):

TombEngine/Scripting/Internal/TEN/Flow/FlowHandler.cpp
```diff
@@ -425,6 +426,13 @@ bool FlowHandler::DoFlow()
 		}
 		else
 		{
+			// Check if there is at least 1 playable level
+			if (g_GameFlow->GetNumLevels() == 1)
```
Lwmte, line 430: I am not sure TEN should close if there is no levels exist on the script but title does. Technically, this is correct scenario, and "crashing" engine in such scenario more resembles TRNG or original engine behaviour. I'd just throw a warning and allow engine to load and play title level anyway.

TombEngine/Game/control/control.cpp
```diff
@@ -271,6 +271,13 @@ unsigned CALLBACK GameMain(void *)
 	else
 		g_Renderer.RenderTitleImage();
 
+	//Check if there is the title in Gameflow.lua
```
Lwmte, line 274: Can you move this code somewhere higher in hierarchy? Lack of any levels in script is a very low-level error, so I suggest moving this to somewhere around `LoadFlowScript`. Proper way of throwing such error would be producing an exception in that function which will feature custom error message. Then TEN will automatically catch it, log it and exit around line 324 in winmain.cpp.

Lwmte (changes requested):

Lwmte (approved):

Sezzary (approved):

PR #1089, opened 2023-04-29
Switch state ambiguity fixes

Lwmte:
This pull request finishes what @Adngel started doing on his February PR #987 .

* Swaps `SwitchStatus` enum entries to reflect real-world states of the switch entity. Before, in Core's logic, `SWITCH_ON` have meant "switch is ready to be pushed", while `SWITCH_OFF` meant "switch was already pushed". In real life, usually you think that switch is ON when it's actually ON, i.e. was pushed.
* ~~Removes ambiguity for the jumpswitch object. Adngel already took care of big button object because it also had flipped states for on/off, but seemingly forgot about the same problem for the jumpswitch object.~~
* Removes some of the magic numbers related to the switches and fixes some formatting.

**NOTE:** `SwitchStatus` enum entries were not flipped in `jump_switch.cpp` file, because jumpswitch already implied flipped entries because of the asset problem which has to be fixed in a similar way to a big button. On contrary, big button `SwitchStatus` enum entries were nevertheless flipped back, because Adngel also did it in his February PR.

**Things to test:**

* ALL switch objects: including buttons, levers, jumpswitches, water switches, valves, reach holes etc.
* Include testing with and without timers
* Include testing with and without activation masks (like in Palace Midas)


Kubsy:
- First time interacting with reach hole it works but then you can interact with it again and will do nothing. Then it is not interactable anymore.
https://user-images.githubusercontent.com/80340234/235315759-c783edbc-d1dd-4232-9e72-f46aacca1831.mp4

- Jumpswitch does not work, neither with unflipped states and flipped states
https://user-images.githubusercontent.com/80340234/235315788-15d07c03-8b59-4a2f-bd79-c920d3ef2266.mp4

- Cannot interact with crowdove switch.
Uploading Desktop 2023.04.29 - 18.19.38.04.mp4…

I did not test timers and activation mask so I will update this comment

Lwmte:
> * First time interacting with reach hole it works but then you can interact with it again and will do nothing. Then it is not interactable anymore.

This is an asset issue, as described in Discord. Second transitional animation lacks `Kill Item` animcommand.

> When you interact with valve, the flame emitter flashes.

Needs more investigation.

Kubsy:
- The timer on the jump switch does not work. I put 3 on timer field and the jump switch from guard.wad (the white jump switch) did not reset and trigger anything with timer being specified.

- I am still not able to trigger stuff with sequence switches. How does it work? Does it only work on sequence door?

Lwmte:
> * The timer on the jump switch does not work. I put 3 on timer field and the jump switch from guard.wad (the white jump switch) did not reset and trigger anything with timer being specified.

It works for me (answered on Discord as well). You need to use an original jumpswitch (since community rejected asset update) which should be two-way (have recovery animation).

>     * I am still not able to trigger stuff with sequence switches. How does it work? Does it only work on sequence door?

Original sequence switches were hardcoded to a particular setup and they miss needed "kill entity" animcommand in their transitional animations. You need to add these animcommands to both transitional animations to make them work.

Kubsy (approved):

PR #1097, opened 2023-05-03
Default input bindings

Lwmte:
Adds "Reset to defaults" menu to control configuration menu.
Additionally, either when resetting to defaults or launching TEN for the first time, automatically binds XBOX mapping instead of keyboard mapping, if XBOX controller is connected. Keyboard mapping remains active in such case too, as it implicitly works as a primary layout.

Things to test:

- Whether key bindings are assigned correctly on first TEN launch (delete `HKCU/Software/TombEngine` in registry to test).
- Whether keyboard layout is correctly reset to defaults.
- Whether XBOX binding is applied when resetting layout to defaults in menu.
- Whether XBOX binding is correct (test with XBOX controller).

### DIFF
--- a/Documentation/Changes.txt
+++ b/Documentation/Changes.txt
@@ -25,6 +25,7 @@ Version 1.0.9
 * Add missing gunflash for some entities, also include dynamic light and smoke to all gunflashes.
 * Add log reports if title level or other levels don't exist.
 * Add better error handling for missing font, sprites or shaders.
+* Add "Reset to defaults" entry to controls menu and automatically bind XBOX gamepad profile if connected.
 
 Lua API changes:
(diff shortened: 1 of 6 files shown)


Kubsy:
Whether key bindings are assigned correctly on first TEN launch (delete HKCU/Software/TombEngine in registry to test). - **Works**

Whether keyboard layout is correctly reset to defaults. - **works**

- Whether XBOX binding is applied when resetting layout to defaults in menu. - **Couldn't get this to work even with xbox controller connected.**

- Whether XBOX binding is correct (test with XBOX controller). - **as above**

Sezzary (approved):

Kubsy (approved):

PR #1109, opened 2023-05-13
Game dir arg

hispidence:


hispidence:
Provides a new command-line option, -gamedir, that lets you specify another path on disk where all game assets are stored.
This includes (meaning these are the folders you must put in the directory pointed to by the path):
Audio
Data
Saves (optional; the game will create the folder here if it doesn't exist)
Screens
Scripts
Shaders
Textures

In addition, the game will create a Logs folder at this path, too.

Note:
dlls must remain in their original directory

todo separately: The pre-build step of copying shaders to the project's target directory should be updated to give the option to copy them to gamedir instead.

A rough idea of how this could be done:
The macro `$(LocalDebuggerCommandArguments)` contains the arguments passed to the game when debugging (i.e. it will include -gamedir and the path, if it's there). We could write a batch script that takes, as an arg, whatever string the macro expands to; it could then xcopy the shaders to the directory if it finds a path in the string, or to `$(TargetDir)/Shaders` if it doesn't. I gave this a go myself but my Windows batch scripting skills suck and I couldn't get something that worked properly.

Lwmte (commented):
I have no objections. Was asset loading (shaders, level files, audio tracks) tested? If yes and working correctly, I think this PR can be merged.

Lwmte (approved):

TombEngine/Game/savegame.cpp
```diff
@@ -1327,14 +1327,14 @@ bool SaveGame::Save(int slot)
 	return true;
 }
 
-void SaveGame::AddGameDirToSavePath(const std::string & dirLocation)
+void SaveGame::AddGameDirToSavePath(const std::string& dirLocation)
```
Sezzary, line 1330: I see both "dir" and "path" used interchangeably in this PR. Why not stick to one? Or is there a specific distinction here? 🤔

hispidence: Usually "path" is where it could be a file rather than a dir, but I used `AddGameDirToSavePath` here because it sounds clearer than `AddGameDirToSaveDir` which makes it sound like some sort of weird filesystem addition is going on.

PR #1128, opened 2023-05-28
Fix and improvements.

TonyTEN1:


Kubsy (approved):

TombEngine/Specific/level.h
```diff
@@ -137,15 +137,16 @@ struct LEVEL
 
 extern std::vector<int> MoveablesIds;
 extern std::vector<int> StaticObjectsIds;
-extern bool IsLevelLoading;
 extern LEVEL g_Level;
 
+inline std::future<bool> level_load_ft;
```
Lwmte, line 142: Can we use different naming convention here? I recommend using PascalCase.

TonyTEN1: Fixed

Lwmte (approved):

PR #1129, opened 2023-06-02
make puzzle_done show up if there is no trigger set

Nemoel-Tomo:


TombEngine/Objects/Generic/puzzles_keys.cpp
```diff
@@ -333,7 +333,21 @@ void PuzzleDone(ItemInfo* item, short itemNumber)
 	auto triggerIndex = GetTriggerIndex(item);
 
 	if (triggerIndex == 0)
+	{
```
Lwmte, line 336: Do not copypaste like this. Copypasting code which is literally repeating several lines down the line is a bad habit.

PR #1130, opened 2023-06-04
Revised slope deflection

Sezzary:
Theoretically improves player deflection when colliding with slope edges. The old method simply checked for a slope above or below the current position and ignored the slope's aspect angle, resulting in funky behaviour with certain setups. This new method does a slightly more advanced geometry assessment to provide more accurate collision.


Lwmte:
I don't have big objections because I don't fully understand the difference between approaches and still waiting for demo video to know the changes :)

Kubsy (approved):

Lwmte (approved):

PR #1151, opened 2023-06-29
Raising block fixes

Lwmte:
* Only shakes it if OCB is not zero
* Simplifies shaking procedure
* Removes abandoned commented code

### DIFF
--- a/TombEngine/Objects/TR5/Object/tr5_raisingblock.cpp
+++ b/TombEngine/Objects/TR5/Object/tr5_raisingblock.cpp
@@ -20,8 +20,9 @@ void InitializeRaisingBlock(short itemNumber)
 	auto* item = &g_Level.Items[itemNumber];
 
 	short roomNumber = item->RoomNumber;
-	FloorInfo* floor = GetFloor(item->Pose.Position.x, item->Pose.Position.y, item->Pose.Position.z, &roomNumber);
-	if(floor->Box != NO_BOX)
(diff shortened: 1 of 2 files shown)


TombEngine/Objects/TR5/Object/tr5_raisingblock.cpp
```diff
@@ -38,6 +39,22 @@ void InitializeRaisingBlock(short itemNumber)
...
+void ShakeRaisingBlock(ItemInfo* item)
+{
+	SoundEffect(SFX_TR4_RAISING_BLOCK, &item->Pose);
+
+	if (item->TriggerFlags == 0)
+		return;
+
+	if ((item->Pose.Position.ToVector3() - Camera.pos.ToVector3()).Length() < BLOCK(10))
+	{
+		if (item->ItemFlags[1] == 64 || item->ItemFlags[1] == 4096)
```
Sezzary, line 51: Should these be bitwise flag checks, or are plain comparisons necessary for correct behaviour?

Lwmte: I have no idea. It seems it is a remnant of the decompiled code from tomb4 or tomb5.

Kubsy (approved):

TombEngine/Objects/TR5/Object/tr5_raisingblock.cpp
Kubsy: these make no sense to me either

Sezzary (approved):

PR #1153, opened 2023-07-06
enhanced horizontal bar system

Nemoel-Tomo:
btw what are all this earlier commits in this PR?

enhanced horizontal bar system:

You need the new Lara object in the wad below for the new animations:

* Keep the action button pressed to make Lara hold the bar.
* press forward to swing, 
* press back or (back and forward) to make Lara turn around.
* if you release the action key while lara is swinging she jumps off.


Nemoel-Tomo:
[horizontal_bar_Lara.zip](https://github.com/MontyTRC89/TombEngine/files/11972406/horizontal_bar_Lara.zip)

here is the wad with the new Lara animations

TombEngine/Game/Lara/lara_objects.cpp
```diff
@@ -258,46 +258,161 @@ void lara_as_pulley(ItemInfo* item, CollisionInfo* coll)
...
+	const auto& barItem = g_Level.Items[player.Context.InteractedItem];
 
 	item->Animation.IsAirborne = true;
 
 	if (item->Animation.FrameNumber == GetAnimData(*item).frameBase)
 	{
-		int distance = 0;
+		int dist = 0;
 		if (item->Pose.Orientation.y == barItem.Pose.Orientation.y)
-			distance = (barItem.TriggerFlags / 100) - 2;
```
Adngel, line 293: I didn't reach to understand the maths for the distance.
What value shall I put in the OCB to make Lara jump 1.5 sector, or 4 sectors?
and
Why does the Lara orientation matters for the else case? Does it really reaches to happens?

I know you didn't touch that but, could this be a chance to consider a new method?

Adngel (commented):
Hi, I've been reading the branch and gossiping the wad.

Now I understand better the value of these changes. The animations looks glitching but that's definitely wad issue. About the code it brings states that with some tweaks can be used to make the system work.

So there are 5 new states:

- LS_HORIZONTAL_BAR_SWING					 128 (old)
- LS_HORIZONTAL_BAR_JUMP					129 (old)
- LS_HORIZONTAL_BAR_SWING_STOP				190
- LS_HORIZONTAL_BAR_IDLE						191
- LS_HORIZONTAL_BAR_SWING_START				192
- LS_HORIZONTAL_BAR_IDLE_TURN_180			193
- LS_HORIZONTAL_BAR_SWING_TURN_180			194

I think we may not need all of them, as some are transitionals and others similar in what player inputs matters.

LS_HORIZONTAL_BAR_SWING_STOP
LS_HORIZONTAL_BAR_SWING_START

These states wouldn't be used for calls, instead the system would call the goal state (BAR IDLE, or BAR SWING) and the animation system would navigate through the intermediary animations to reach there. 

While those transition animations are happening, these states could be running on them, but actually, they wouldn't be doing anything that were not done by the main BAR IDLE and BAR SWING.

Extra features from IDLE and SWING (like ability to turn 180º or can release the bar) wouldn't affect the transition (they will keep their way until "Next Anim" or the "State Changes" allows it to do the require animation). So I think these states are not really needed.

If decide to keep a state for transitional animations, (maybe because have any idea planned for the future that could expand BAR SWING and BAR IDLE behaviours), I think these transitional states could be combined into 1 and still would work.

LS_HORIZONTAL_BAR_IDLE_TURN_180
LS_HORIZONTAL_BAR_SWING_TURN_180

I also think these two states could be combined into 1, their difference is that IDLE TURN allows to update when the bar is release, but still, that won't have any effect until the turning animation ends and reaches to the idle animation controlled by BAR IDLE (that also will have the check of bar release)

And if player drops the action key while is doing the SWING TURN, the animation system still will complete the animation and take the exit dictated by BAR SWING state.

I see the turn state could be useful to indicate the activation of that particular skill (do the 180º turning), but doing it on idle or while rolling around the bar, is the same.

In your wad there are two turnings, but one can only happen on IDLE and the other only can happens on SWING, they can't happens both at the same time so it can use the same State for both.

In my case I'm considering three turning animations, but still for the three can be managed with the same state. (One state to calls them and control there not more behaviour affecting until their completion). 

Excepting that, I haven't got any other issue with this PR, I've tested and it works, the glitchy issues and wrong animation transitions that I've experienced, are caused by the WAD and how the animations are set  in there, but not by the code here.

PR #1157, opened 2023-07-09
new DrawNearbyTileFlags

Adngel:
An update to the data showed in the debug page LOGIC. (To see it, must be executed in debug mode and use the keys F10-F11 to navigate along the pages).

This PR makes 4 changes.

1.  Creates a new function to paint colored floating text on the nearby tiles, indicating the floordata flags those sectors have.
- Stopper
- Death
- Monkeyswing
- Bettler (MinecartRight)
- Trigger Triggerer (MinecartLeft)
- MinecartStop

2. Updates the pathfinding drawing function, so now if the phatfinding box where Lara is located, is a blockable box, (the one marked with the grey button in the TombEditor), with this change, that box will be painted in green or red depending of its state, instead of the default cyan. (For example if there is a door in there, opening and closing the door may change the state of that pathfinding box).

3. In Lara.pp code, it has been added some lines to restringe the activation of these drawing functions, to happen only in DebugMode.

4. Adds a new AddDebugString, it requires an extra value, the Debug page where the text must be drawn. If the correct debug page is being rendered, then the process will continue with the normal AddString function. Using this method tries to make it more similar to other used functions like AddDebugSphere or AddDebugBox.

The purpose of these additions are to make easier the diagnosis and test of other features involved with the floor data flags. (Like pushables objects, shatters or electric cleaner).


Kubsy (approved):

TombEngine/Game/collision/floordata.cpp
```diff
@@ -924,4 +926,76 @@ namespace TEN::Collision::Floordata
...
+		{
+			for (int z = minZ; z < maxZ; z++)
+			{
+				point.x = BLOCK(x);
+				point.z = BLOCK(z);
+
+				pointColl = GetCollision(point);
+				point.y = pointColl.Position.Floor;
+
+				if (pointColl.Block->Stopper)
```
Kubsy, line 957: would this be better in a switch-case statement?

Kubsy (approved):

TombEngine/Game/collision/floordata.cpp
Adngel: The issue is that a sector may have multiples flags activated at the same time. (monkey + stopper for example) so it would draw two sphers (the red and orange one).

If I put it like a switch, it would draw only 1, the first that it find the condition.

Kubsy: Ah ok I understand :)

Lwmte:
Please consider drawing block attributes as boxes, not as spheres.

Adngel:
> Please consider drawing block attributes as boxes, not as spheres.

At the end I discarded the geometry spheres, with Sezz's help, I included floating text labels, that I think will be more suitable to provide the information of the block's flags.

In this last update I have reworked the function a bit to reduced repetition and I added AddDebugString, it calls the Sezz's AddString but checking first with the Debug's page like the AddDebugSphere or AddDebugBox do.

Sezzary (approved):

Sezzary:
Looking good! I've removed my string constants, as I realise I'm often excessive with them. Here it's fine to leave strings inlined imo.

PR #1159, opened 2023-07-15
Bindable vehicle actions

Sezzary:
Input refactor.

- Introduce bindable input actions for vehicles.
- Organise default bindings into vectors.

To test:
- Vehicle controls, particularly dismounts.
- Setting and saving input action binding configurations.


.gitignore
```diff
@@ -23,3 +23,4 @@ packages/
 *.str
 *.ten
 Logs/
+/Documentation/Changes.txt
```
Kubsy, line 26: oop don't put that to gitignore :(

PR #1162, opened 2023-07-16
Object handler

Lwmte:
Introduces object handler wrapper class which implements a safeguard measure to prevent accessing unavailable object IDs (e.g. NO_OBJECT or bigger than highest available object ID).

This should help to alleviate problems with x64 build which exposed global systemic issue with unsafe memory access to `Objects` array which was apparently masked by x86 compiler which tolerates that, unlike x64 one.


TombEngine/Game/Setup.h
```diff
@@ -142,7 +142,49 @@ struct ObjectInfo
 	}
 };
 
-struct STATIC_INFO
+class ObjectHandler
+{
+	private:
+		ObjectInfo m_objects[ID_NUMBER_OBJECTS];
```
Sezzary, line 148: TEN conventions use PascalCase for member variables. The renderer and scripting system are currently exceptions since they were created before any hard rules were set.

Sezzary (approved):

PR #1078, opened 2023-04-21
Better look camera

Sezzary:
Significantly improves look camera functionality.

To test:
- Looking around in various player states.

PR #1164, opened 2023-07-19
Remove legacy input bit fields

Sezzary:
Remove `TrInput` and `DbInput` once and for all.
To be merged only *after* #1078 and *after* 1.1.0 release.

When reviewing, ensure these conversions have been done correctly with the right logic preserved:
- `TrInput & IN_A` -> `IsHeld(In::A)`
- `DbInput & IN_A` -> `IsClicked(In::A)`
- `TrInput & (IN_A | IN_B)` -> `IsHeld(In::A) || IsHeld(In::B)`
- Etc.


Adngel (approved):
I have not seen anything out of place in the codes.

I've tested playing my Lara's gym, where I could make all the Lara's movements with not problem.
I've tested also all the switches, puzzles, doors, searchable objects and pushable. I neither had any issue.

PR #1167, opened 2023-07-23
Vec2 scripting object refactor

Sezzary:
Refactors `Vec2` to be float-based and adds missing multiplication and division operators.

To test:
- Saving/loading `Vec2`.
- Multiplying two `Vec2` objects.
- Dividing a `Vec2` by a number.
- Confirm `Vec2` is in fact float-based.

Previously the PR also introduced an integer-based `Vec2i`, but it's probably unnecessary since all Lua numbers are implicitly double-precision floats.


TombEngine/Scripting/Internal/TEN/Logic/LogicHandler.cpp
```diff
@@ -634,9 +642,9 @@ void LogicHandler::GetVariables(std::vector<SavedVar>& vars)
 
 				case sol::type::userdata:
 				{
```
Adngel, line 644: I'm not sure, but shouldn't be here another if case for the vec2 float?

Adngel (commented):
As you know, the test I did with lua code, I could create and use both vec2 and vec2i, but I couldn't save-load the variables saved in vec2 (float version).

TombEngine/Scripting/Internal/TEN/Logic/LogicHandler.cpp
Sezzary: Thanks, this might be why it wasn't loading.

Adngel (approved):
After test it, the new format type already manages to save and load in my test level.

I did not see any other issue or concern about the code.

TombEngine/Game/savegame.cpp
```diff
@@ -113,7 +118,7 @@ Save::Vector3 FromVector3(const Vector3& vec)
 	return Save::Vector3(vec.x, vec.y, vec.z);
 }
 
-Save::Vector3 FromVector3(const Vector3i& vec)
+Save::Vector3 FromVector3i(const Vector3i& vec)
```
Lwmte, line 121: I don't fully understand why you have converted this one from `Vector3` to `Vector3i` while all other changes seemingly do the opposite (removing `i` everywhere). Can you elaborate?

Sezzary: Previously, `FromVector3()` had two overloads, but this made it unclear which type was being converted. Instead, now there are two versions: `FromVector3()` and `FromVector3i()`.

PR #1176, opened 2023-08-02
Adngel pushables remake

Adngel:
This branch is updating the codes of the Pushables Objects.

Is not totally finished, but at the moment I'll give it pause and won't add more to this branch (which is already quite long). I make PR to invite for tests, although myself will keep doing tests also along the week.

/----------------------------------------------------------------/
### Content
- Code overhaul. Although most of the original concepts from the old code remains, this code has been modularized in several files, to make them more readable and in case of need to change some element, help to make it more encapsulated. (So if Bridge functions get refactored someday, these can change the PushableObjects_BridgeCol code, and the rest should go on fine).
Also there were tons of optimizations, deleted redundancies, and created some structures to be shared by all pushables (instead of have each pushable with a copy of the same sound and animation data).
- Separation between **Pushable_Objects** and **Climbable_Pushable_Objects**.

1. **Pushable_Objects** are the classic pushables from TR4 that uses Object collider.
2. **Climbable_Pushable_Objects** are the pushables that uses the Bridge collider to allow Lara and other objects walk over it. The collision bounding box is used to indicate the sizes of the Bridge collider.

- Several bug fixes.
- Creation of a State machine. Now there are new states for the pushables (some related to falling and water rooms). So instead to have them in a long switch with cases, it has been implemented each one on their own function.
- Sounds Map to stores pushables sounds per ground material.
- New OCBs.
  - 0 Standard behaviour.
  - +1 Can Fall, If added, Lara will can push the pushable through border edges.
  - +2 Don't Align Center, If added, Lara'll be able to grab the pushable without need to positionate in the middle of the tile.
  - +4 Is Buoyant, If added, the pushable will float to the surface when it were in a water room.
  - +8 Animation System, If added, Lara will use the classic TR1-TR3 push pull animations. (Need WAD update).

- Pushables now can be used with Raising Blocks. (However stacked pushables may present issues, specially if both are Climbable pushables).

### Known Issues
- When Stacked pushables are moved vertically (Raising Block, twoBlock Platforms, trapdoors, etc), they may loose the stack link. (Causing the top one to fall on the ground)
- When a pushable is moved over other pushables, this removes the stopper flag from the ground. (What may affect other entities behaviour).

### Resources
WAD with Lara classic animations:
v.1: deleted
v.2: deleted
v.3: deleted
v4: deleted
v5: (Copy of the main TEN WAD) : https://drive.google.com/file/d/1TLu8mxhAWDzVu69CQIMmAcw2IQwo6kw_/view?usp=sharing

XML with the new sounds: (pending)
https://drive.google.com/file/d/1QhiQV7PZCXyc9tKEiugyFNS60zn5wekb/view?usp=sharing


Stranger1992:
Testing - Sunday 20th August:

**- **Room used to test:** 
- image**

image

**- Bugs and issues**


TEN-20-08-2023-17-14-07
- > Only allowed me to throw off pushables (OCB 5) that were *2* and *4* clicks above the water.
TEN-20-08-2023-17-19-30
- > Lara does not collide if swimming directly up towards the pushable. This be an issue with the asset used here.
TEN-20-08-2023-17-18-34
- > Lara still comes too far forward when pushing which causes her to fall off the end of the platform in all cases.

Stranger1992:
Level Design Change:

image

TEN-20-08-2023-17-36-06
If the ground is flat: ALL OCB 8 pushables fell ✅

Adngel:
> Level Design Change:
> 
> image
> TEN-20-08-2023-17-36-06 If the ground is flat: ALL OCB 8 pushables fell ✅

Ohm, thank you.

Yes, I intentionally blocked the ability to push pushables on edges if the next sector is not flat, In the future, I may try implement pushables sliding on slopes, so I may modify that restriction.

However, this this particular case you showed, is true if pushables are floating there shouldn't be any reason to forbid the player throw them because the floor isn't flat. I'll try polish that.

Adngel:
Wolf Animations and States checked. I removed the TODO comment.

Stranger1992:
First of all: I think you've done amazing work with this pushable branch I really do!

However as testing: its my job to break it. One bug I have is if lara throws a pushable off an edge , on to a raising block, and then raises that up: she can no longer push or pull the block.

https://github.com/MontyTRC89/TombEngine/assets/84292688/f2ecec76-143f-447e-b050-93ff462d03a1

Stranger1992:
> First of all: I think you've done amazing work with this pushable branch I really do!
> 
> However as testing: its my job to break it. One bug I have is if lara throws a pushable off an edge , on to a raising block, and then raises that up: she can no longer push or pull the block.
> 
>  2023-09-03_15-25-11.mp4

This is the case for both TR4 and TR1-TR3 animations

Stranger1992:
✅ Setting a raising block as climbable in the editor works as expected . Once raised: she can climb up it and any pushable blocks which are there too

Stranger1992:
"Normal" pushables are allowed to be pushed and pulled off raising blocks after being thrown: however they cannot be pushed off a raising block again for a second time. See below for link to video.

https://discord.com/channels/1072511394735464498/1072569001986773023/1147903123843711077

Stranger1992:
> "Normal" pushables are allowed to be pushed and pulled off raising blocks after being thrown: however they cannot be pushed off a raising block again for a second time. See below for link to video.
> 
> https://discord.com/channels/1072511394735464498/1072569001986773023/1147903123843711077

This is also the case if lara tries to push it off a normal ledge for a 2nd time:

https://github.com/MontyTRC89/TombEngine/assets/84292688/208efeb8-2ef4-4f90-b32c-1ac9b8ce3124

Stranger1992:
Pushable cannot be pushed under a bridge IF on top of a bridge already

Stranger1992:
> 
> 
> Pushable cannot be pushed under a bridge IF on top of a bridge already



Or if under a bridge at all.

Stranger1992:
> > 
> > Pushable cannot be pushed under a bridge IF on top of a bridge already
> 
> Or if under a bridge at all. 
> 
> 

Also the case for Climbable_pushables

Stranger1992:
> > > 
> > > Pushable cannot be pushed under a bridge IF on top of a bridge already
> > 
> > 
> > Or if under a bridge at all. 
> > 
> 
> Also the case for Climbable_pushables 

Also is not due to collision (ie: the block is 4 clicks high and the old bridge was directly touching it. If I raise the bridge up 1 click it still happens:

Stranger1992:
These bridge issues may not be related to your branch btw as I tested walking lara under bridges and any bridge 4 clicks or below lara cannot run under (or 3 if she is crawling)

TombEngine/Objects/Generic/generic_objects.cpp
```diff
@@ -97,6 +98,12 @@ static void StartObject(ObjectInfo* object)
 		}
 	}
 
+	for (int objectNumber = ID_PUSHABLE_OBJECT1; objectNumber <= ID_PUSHABLE_OBJECT10; objectNumber++)
+		InitPushableObject(object, objectNumber);
+
```
Stranger1992, line 103: Why does this part of your code not follow suit with other objects? For other generic objects you can set things like the hit effect or shadow type but not with this?

Kubsy:
https://github.com/MontyTRC89/TombEngine/assets/80340234/7e078df8-9c67-4a9d-86e3-de679d312de8

lara teleports beneath the pushable


https://github.com/MontyTRC89/TombEngine/assets/80340234/678d8753-4ea5-48a3-a990-cb0785b6f770

lara refuses to gram certain sides after flipmapping

Kubsy:
[level.zip](https://github.com/MontyTRC89/TombEngine/files/12536284/level.zip)

TombEngine/Objects/Generic/generic_objects.cpp
Adngel: I also saw a bit weird, if you follow those initpushableobject functions, you will eventually reach to another initialization block that similar to the other objects. I don't know why it was separated in that way, I didn't notice anything "so" special, but well, it didn't hurt so I kept it.

(It looks "added" because I moved it from TR5 objects into the general objects).

Adngel:
I tried to review these bugs but I hadn't got luck:

1. One bug I have is if lara throws a pushable off an edge , on to a raising block, and then raises that up: she can no longer push or pull the block.
2. "Normal" pushables are allowed to be pushed and pulled off raising blocks after being thrown: however they cannot be pushed off a raising block again for a second time. 
3. Pushable cannot be pushed under a bridge IF on top of a bridge already...
4. Lara teleports beneath the pushable (When climbs up from water)
5. Lara refuses to grab certain sides after flipmapping

I couldn't replicate 1, 2, 3 nor 5. (May be an asset issue? review the collider boxes of the objects (pushables and bridges).

About the 4. I've increase a value so Lara will detect the bridge floor at a higher height, what will reduce the times she appears inside the block (what was making her to teleport under the pushable), however is not an ideal solution (as after she climbs, she tends to fall), hopefully in a future when the floor data get reviewed, this aspect could get polished.


(Kubsy, I'm afraid I could not use your test level, the code has change the animations locations so I can't push/pull anything in that level anymore).

TombEngine/Objects/Generic/Object/Pushables/Context.h
```diff
@@ -4,13 +4,13 @@ class Vector3i;
...
 	{
 		Air,
-		Ground,
-		Slope,
-		Water,
-		GroundWater
+		FlatFloor,
+		SlopedFloor,
+		DeepWater,
+		ShallowWater
```
Adngel, line 13: I think these names:

DeepWater and ShallowWater are unsuitable, as may lead to a concepts confusion.

Usually I understand DeepWater to the water rooms that have a depth of 3 clicks or more. And Shallow Water to the ones that have a depth of 2 clicks or less (usually used to identify the parts where Lara can swim and where Lara can wade).

However this context function is not checking that. What it needs to see is if the block is in the water room floor, or is higher (so is sinking or floating).

Is something similar like the flat floor or on air states, but for water rooms. ( their own version because are very frequent cases so I could use those types instead of check is on the ground, Is it also in water room?, so the GroundWater state implicit both conditions of that situation.

Formally, I also was using a ShallowWater type so I had 3, DeepWater, ShallowWater and GroundWater) but reviewing  the code, I found that the behvarious that differ DeepWater from ShallowWater were mostly anecdotal, so I combined those cases into only one (Water), and did the distinction in the states functions (that were only in a couple of moments).

Stranger1992:
Hi,

Ive just tested this branch again and the pushable sound is playing , on loop even when not interacted with.

https://github.com/MontyTRC89/TombEngine/assets/84292688/6786f7dc-495c-40dc-8b6f-02f85f8cf6ed

Stranger1992:
The negative OCB works nicely though to prevent further movement of the pushable.

PR #1177, opened 2023-08-06
Remove jeep reverse toggle

Sezzary:
To test:
- Animations while driving in a jeep.

### DIFF
--- a/TombEngine/Objects/TR4/Vehicles/jeep.cpp
+++ b/TombEngine/Objects/TR4/Vehicles/jeep.cpp
@@ -64,14 +64,14 @@ namespace TEN::Entities::Vehicles
 		JS_CRASH_BACK = 3,
 		JS_CRASH_LEFT = 4,
 		JS_CRASH_FORWARD = 5,
-		JS_BRAKE = 6,//?
-		JS_FWD_LEFT = 7,//FWD = forwards
+		JS_BRAKE = 6,
+		JS_FWD_LEFT = 7,
 		JS_FWD_RIGHT = 8,
 		JS_MOUNT = 9,
 		JS_DISMOUNT = 10,
 		JS_JUMP = 11,
 		JS_LAND = 12,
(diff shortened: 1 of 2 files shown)


Adngel (commented):
Testing this branch, I got an issue with Lara's animation.

When I press DOWN to go back, Lara's start to rotate the head to look back (animation 23), but then goes to rotate forward (animation 25). She does that continuosly all the time while I keep pressing the DOWN key.

To solve that, I understand it would be possible do changes to the TEN Wad, in the states changes from animation 42 (although I don't know if is failing because the Animation State is 0 or 10). Or maybe there is something else in the code that aims to manage this issue.

Adngel:
About some more personal feedback, I'm not too sold on the removal of the rear drive feature, I guess part of that may be due to nostalgia. But I would like to share my point of view.

I understand that the new ability to reverse will be useful to control the Jeep with more freedom and helps you get out when the jeep gets stuck, more quickly and naturally.

However it restricts vision. This new way of moving backwards leaves the camera too close so it's hard to see when the Jeep is approaching an edge, or unable to show where obstacles or enemies may be while driving backwards).

The traditional method of turning the controls, may be more clunky, but it provided a way to drive in such direction with a better camera an a good range of animations to sort that movement. Rear drive was not the most important feature of the Jeep since a gameplay perspective, but I think is one of those unique features that support the veicle identity and differenciate it from the others.

So would suggest, don't replace one with another, but make both available. 
- Perhaps the new backwards by DOWN key could persist (although I recommend move the camera a bit further so it were possible to see the full jeep on the screen).
- And the old method, could be moved to the ROLL key, what could make more sense than the original WALK key.

I think the new method will welcome and mostly used, but if someone still want to use the classic method (for personal taste, nostalgy or necesity in an specific scenario, it could keep using it). Besides, all animations and states are already implemented, so I think would be a shame they get blocked and unused by code.

At least, that's how I see it.

PR #1181, opened 2023-08-07
Mouse input base

Sezzary:
Adds base mouse functionality we can use in future GUI developments.

Main changes:
- Add mouse polling.
- Allow mouse inputs to be bound in control settings.
- Add Mouse Sensitivity and Mouse Smoothing settings and registry keys.
- Add `GetCursorDisplayPosition()` script function.
- Store analog axis values in `Vector2` containers.

NOTE: Commit history isn't accurate, so a squash merge will be best here.


TombEngine/Specific/configuration.h
```diff
@@ -36,15 +36,24 @@ constexpr auto REGKEY_ENABLE_AUTO_TARGETING	   = "EnableAutoTargeting";
...
+	static constexpr auto DEFAULT_MOUSE_SMOOTHING	= 1;
+
 	// Graphics
 	int		   ScreenWidth		  = 0;
 	int		   ScreenHeight		  = 0;
 	bool	   EnableWindowedMode = false;
 	ShadowMode ShadowType		  = ShadowMode::None;
-	int		   ShadowMapSize	  = 1024;
-	int		   ShadowBlobsMax	  = 16;
+	int		   ShadowMapSize	  = DEFAULT_SHADOW_MAP_SIZE;
```
Adngel, line 55: I'm not sure if this is gonna affect, but I'm seeing in this header code (configuration.h) DEFAULT_SHADOW_MAP_SIZE has the value of 1024, however in the source code (configuration.cpp) was taken to replace the shadowMapSize constant that was 512.

Was an intentional change? or would be okey having 512 changed to 1024? Have not tested yet.

Adngel: Yeah, I think there's not gonna be problem with that, I pointed that just to be aware.

Adngel:
Hi!, I've tested and I've got three issues.

1. When I move the mouse, the debug screen shows me the position goes X 0-800 and Y 0-600, although my resolution is 1920x1080. The location of the numbers looked correct ([0,0] top left corner and [800,600] the bottom right corner) But I suposse you may want to review that, to show the range according to the resolution, or to show a percentadge, that could work too.
2. When I move the mouse, I see the axis getting the raw readings, I think it would be useful get also a normalized ones (float values that gets the range clamped betwen 1 to -1), generally is the practice I'm used to, so later I use a custom speed variable to control the mouse speed and smoothness. Unity and Unreal used to provide both (the raw and normalized) So maybe could add a getter for that too?
3. I tested the release and debug mode, but the 4 times I put the game, in all of them it started with the config window (the one to choose resolution and graphics options), I saw strange it pops up always when I put the game. Is it me alone? Could be related to any change of this branch?

Apart of that, all looked to work as intended, I think is a handy and great feature.

Sezzary:
> 

1. 800x600 is the internal screen space resolution, so everything is normalised to it.
2. I don't know how it should work, but I can look into it.
3. Registry is being read incorrectly. That should be fixed.

Kubsy (approved):

PR #1183, opened 2023-08-08
Adngel crumbling platform

Adngel:
I wrote some code for crumbling platform object.

It has 3 ocb functions.

- OCB == 0: Default behaviour. The platform will start to shake when Lara steps in and will crumble during 35 frames (The animation's leght).
- OCB > 0: The platform will start to shake when Lara steps  and will crumble during the number of frames indicated in the OCB.
- OCB < 0: The platform will need be activated with a trigger, then it will crumble during the number of frames indicated in the OCB.

Right now it needs some changes to the object in the WAD.

- The crumbling animation by default is connecting to the falling animation when it ends, that unables the platform to keep crumbling for more than 35 frames. So this version has changed that to allow the crumbling animation loop.
- The original bounding box doesn't cover the whole sector what interrupts Lara's movements on sector edges, this version has changed the bounding box to cover the 511 units in all the horizontal axis during the Idle and Crumbling animations.
- The original platform, had a height offset of 512 units, the code has been prepared to deal with this case, but still I removed that offset in this version, so now the platform will be on the same click height that indicates Tomb Editor.
- This version has fixed the sounds ID in the falling and landing animations.

Crumbling Platform WAD: https://drive.google.com/file/d/18s56vn45Tn9DYU1S_COViU-031v3uM1Q/view?usp=sharing

Without those changes in the wad, the object still works but with some issues that could affect the gameplay.

### DIFF
--- a/TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
+++ b/TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
@@ -102,17 +102,19 @@ namespace TEN::Entities::Traps
 
 			case CRUMBLING_PLATFORM_STATE_FALL:
 			{
+				short& fallVel = item.ItemFlags[1];
+
 				auto pointColl = GetCollision(item);
 				int relFloorHeight = item.Pose.Position.y - pointColl.Position.Floor;
 
 				// Airborne.
-				if (relFloorHeight < 0)
+				if (relFloorHeight < -fallVel)
 				{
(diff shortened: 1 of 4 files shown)


Sezzary (approved):

Adngel:
Mental note, please remember me to add this to the changes.txt when it were ready for merge.

Kubsy:
Will tell dan to update the site

Kubsy (approved):

PR #1196, opened 2023-08-16
Adngel enemies Peru fixes

Adngel:
Some fixes in the Wolf and Bear.

### Wolf
- Now when the wolf does stalking bite attack, it will hurt Lara.
- Wolf now will align to the floor surface

### Bear
- Renamed the states, animations and constants to make code clearer.
- Fixed logical errors: Now bear won't get stuck during the high stance states.
- Fixed an error that caused the bear getting stuck in low ceiling rooms when it used the high stance.
- Now when the bear gets Escape mood, it will run away instead of walk.
- Bear now will align to the floor surface

I checked the error with the Lara's TRex death animation, but happens to be a wad issue, not a code problem, so at the end I didn't edit anything from the TRex.


Kubsy (approved):

Adngel:
Wolf Animations and States checked. I removed the TODO comment.

PR #1199, opened 2023-08-16
Adngel lua weapons functions

Adngel:
- Added the Lua function GetAmmoType
- Added the enumerator LaraAmmoType for the engine uses.
- Added the enumerator AMMO_TYPE_IDS for the Lua api.
- Function AddCallback now will print information in the log if the function is already added, to avoid duplication.
- Fix error in Flow.Inventory builder, SetAction now accepts compound action values like for example:  
     _ItemAction.USE | ItemAction.COMBINABLE_  (But max 3, if there are 4 the game crashes when tries render the object in inventory).


Stranger1992:
image

- ✅Additional options for puzzle items works (`ItemAction.COMBINE` will only show on items that have two parts which is correct behaviour. 
- ❌Zooming in and out with binoculars / lasersight does not work (this was already the case on the `develop` branch but it makes sense for this to be addressed on this PR ?

PR #1204, opened 2023-08-30
Split strings file

Lwmte:
Splits Strings.lua to:

1) SystemStrings.lua - mandatory file which contains all system strings
2) Strings.lua - for user stirngs, such as level names, item names, weapon names, etc.

If Strings.lua contains same entry as was previously loaded from StringsSystem.lua, it will be overwritten. This way we maintain backwards compatibility with Strings.lua files which were already modded by users (there were already complaints on that).


TombEngine/Scripting/Internal/TEN/Flow/FlowHandler.cpp
```diff
@@ -229,14 +229,22 @@ void FlowHandler::SetGameDir(const std::string& assetDir)
...
 void FlowHandler::SetStrings(sol::nested<std::unordered_map<std::string, std::vector<std::string>>>&& src)
 {
-	m_translationsMap = std::move(src);
+	if (m_translationsMap.empty())
+	{
+		m_translationsMap = std::move(src);
+	}
+	else
+	{
+		for (auto& s : src.value())
```
Sezzary, line 245: What is `s`? Maybe we could have a more descriptive name for clarity.

Stranger1992:
Tested with new strings and systemstrings file : worked as expected. Approved.

Stranger1992 (approved):

PR #1208, opened 2023-09-07
Batching + Lua function for display sprites

Sezzary:


Kubsy (approved):

Lwmte (approved):

TombEngine/Game/Lara/lara.cpp
```diff
@@ -441,10 +441,15 @@ std::function<LaraRoutineFunction> lara_collision_routines[NUM_LARA_STATES + 1]
...
 };
 
+#include "Game/effects/ScreenSprite.h"
+using namespace TEN::Effects::ScreenSprite;
+
 void LaraControl(ItemInfo* item, CollisionInfo* coll)
 {
 	auto* lara = GetLaraInfo(item);
 
+	//AddScreenSprite(ID_DEFAULT_SPRITES, 0, Vector2(400, 300), 0, Vector2::One, Vector4::One, 0, BLEND_MODES::BLENDMODE_ALPHABLEND);
```
Lwmte, line 451: I guess this line was here for a test? Please remove it before merging.

TombEngine/Game/control/control.cpp
```diff
@@ -145,6 +147,8 @@ GameStatus ControlPhase(int numFrames)
 
 	for (framesCount += numFrames; framesCount > 0; framesCount -= 2)
 	{
+		ClearScreenSprites();
```
Lwmte, line 150: Why this call is here? Clearing of screen sprites should happen around cleaning up all other renderer objects, such as strings and particles.

Lwmte (changes requested):
Please also rename ALL occurences of `ScreenSprite` to `DisplaySprite`. All 2D space objects must be called `Display...`, not `Screen...`. We already have `DisplayString` and `DisplayPickup`.

Adngel:
I tested the beetles swarms, but I got more bugs with this branch. 

Little Beetles have the next OCBs:

0 - 128 = Total number of scarab you want
1000 += Scarab appear from the floor
2000 += Scarab appear from the ceiling
4000 += Scarab slow release followed by a gush

(Use Flipeffect 31 to make them dissapear).

In develop, the game crashes, but beetle spawn animation worked correctly in these 4 ocb states. However in this branch, they do not. When trigger is activated, it takes some seconds for the beetles swarms appear (although they may be reducing Lara's health points from before). When they do appear, they do not perform any of the spawn animations, they just appear the full swarm at the instant.

Also, using the different bettles, I can see that only the first one appears. (I think they may reach the limit) but still there are "invisible" bettles in these cases.

This is a test level I was testing different beetles:
 https://drive.google.com/file/d/1t7iH08BkgUGfEVejGFoMFJB9jFx28cAv/view?usp=sharing 
 (Each one spawn 128 bettles but with different OCB).

PR #1210, opened 2023-09-07
add lua function GetDisplayPosition()

davidmarr:
Get the 2d position from a 3D position (Vec3) of an object


Sezzary:
Addressed by #1208.

TombEngine/Scripting/Internal/TEN/Misc/Miscellaneous.cpp
```diff
@@ -393,6 +396,31 @@ namespace Misc
...
+	//@usage 
+	//
+	// Example: display string at an object's position.
+	// local string = DisplayString('Example', 0, 0, Color(255, 255, 255), false)
+	// local pos2D = Get2DPosition(Lara:GetPosition())
+	// string:SetPosition(PercentToScreen(pos2D.x, pos2D.y))
+	static Vec2 Get2DPosition(const Vec3& pos)
+	{
+		// TODO: Move to Renderer11Enums.h.
+		constexpr auto INVALID_2D_POSITION = Vector2(FLT_MAX);
```
Lwmte, line 411: I think that instead of returning nonsensical 2D position, you should also return `nil` to Lua in such case. That is, make `Get2DPosition` lua method to return `optional` too.

TombEngine/Scripting/Internal/TEN/Misc/Miscellaneous.cpp
```diff
@@ -393,6 +396,26 @@ namespace Misc
...
+	//@return Vec2 pos2D Projected 2D screen position. NOTE: Screen space resolution is 100x100.
+	//@usage 
+	//
+	// Example: Display a string at the player's position.
+	// local string = DisplayString('Example', 0, 0, Color(255, 255, 255), false)
+	// local pos2D = Get2DPosition(Lara:GetPosition())
+	// string:SetPosition(PercentToScreen(pos2D.x, pos2D.y))
+	static std::optional<Vec2> Get2DPosition(const Vec3& pos3D)
+	{
+		auto pos2D = g_Renderer.Get2DPosition(Vector3(pos3D.x, pos3D.y, pos3D.z));
```
Lwmte, line 410: Maybe instead of reconstructing `Vector3` in place, make an operator which implicitly converts `Vec3` to `Vector3`?

PR #1211, opened 2023-09-07
PointCollisionData class

Sezzary:
Introduces the new and improved `PointCollisionData` class to replace `CollisionResult`. The advantage over the old method is that it fetches and caches computationally expensive data only when needed. It can be easily wrapped for the eventual Lua collision detection object. Additionally, it can theoretically plug in with any potential future collision overhaul.

Usage example:
`auto pointColl = GetPointCollision(pos, roomNumber);`
`int floorHeight = pointColl.GetFloorHeight();`

To get a broad overview of this PR, look at `Game/collision/Point.h` first.

Some general improvements:
- Added the ability to retrieve top sectors.
- Added the ability to retrieve ceiling bridge item numbers.
- Added the ability to query water height values.
- Provided the option to test environment flags.
- Bridges are recognised when fetching floor or ceiling normals.
- Bridges can be wobbled (but not too much).
- etc.


Sezzary:
Postponed until further low-level floordata refactors are complete.

Sezzary:
Full adoption of new functions complete.

Jakub768 (approved):

TombEngine/Game/Lara/PlayerContextData.h
```diff
@@ -8,9 +8,9 @@ namespace TEN::Entities::Player
 		int	  LowerFloorBound = 0;
 		int	  UpperFloorBound = 0;
 
-		bool TestSlipperySlopeBelow = true;
-		bool TestSlipperySlopeAbove = true;
-		bool TestDeathFloor			= true;
+		bool TestIllegalFloorBelow = true;
```
Lwmte, line 11: Why slidable slope is now called illegal floor? In TE and TRLE terminology, "illegal slope" means a slidable slope which can get player into softlock or buggy motion. I suggest to avoid using word "illegal" in this context.

TombEngine/Objects/Generic/Object/Pushable/PushableCollision.cpp
```diff
@@ -28,24 +30,24 @@ namespace TEN::Entities::Generic
 		if (pushableItem.Status == ITEM_INVISIBLE || pushableItem.TriggerFlags < 0)
 			return false;
 
-		auto pointColl = CollisionResult{};
+		auto pointColl = GetPointCollision(pushableItem);
```
Lwmte, line 33: Why you are doing `GetPointCollision()` here, and immediately down the line, inside else block?

TombEngine/Game/collision/collide_room.cpp
```diff
@@ -943,11 +823,11 @@ short GetNearestLedgeAngle(ItemInfo* item, CollisionInfo* coll, float& distance)
 
 			// Get front floor block
 			auto room = GetRoomVector(item->Location, Vector3i(ffpX, y, ffpZ)).RoomNumber;
-			auto block = GetCollision(ffpX, y, ffpZ, room).Block;
+			auto sectorPtr = &GetPointCollision(Vector3i(ffpX, y, ffpZ), room).GetSector();
```
Lwmte, line 826: I am against using `Ptr` suffix everywhere. We can already tell the data type by just peeking definition or seeing whether `.` or `->` is applicable in every case.

Jakub768 (approved):

PR #1226, opened 2023-10-02
add constant "unlimited" in SystemStrings.lua

davidmarr:
Add constant STRING_UNLIMITED to be able to translate the word "unlimited"

### DIFF
--- a/TombEngine/Game/gui.cpp
+++ b/TombEngine/Game/gui.cpp
@@ -2710,7 +2710,7 @@ namespace TEN::Gui
 
 					if (AmmoObjectList[n].Amount == -1)
 					{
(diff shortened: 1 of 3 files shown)


TombEngine/Game/gui.cpp
```diff
@@ -3033,7 +3033,7 @@ namespace TEN::Gui
 						{
 							if (numItems == -1)
 							{
-								sprintf(textBuffer, "Unlimited %s", g_GameFlow->GetString(invObject.ObjectName));
```
Lwmte, line 3036: Why do you use `strcat` and `strdup` here? Just make `%$` mask part of the lua string. User may want to appear ammo name prior to "unlimited", or it may be language-specific.

Lwmte (changes requested):

Lwmte (approved):

PR #1241, opened 2023-10-24
Delete SetTitleScreenImagePath references

Kubsy:
Aim of the pr is to napalm SetTitleScreenImagePath  references throughout the code because:

A) It won't be implemented anytime soon
B) you draw a 1920x1080 (up to 2048x2048) sprite which can fill the whole screen space area for title
   - Also means users won't be doing the usual "delete files" procedures as they do with exes etc, providing better security.

### DIFF
--- a/Scripts/Gameflow.lua
+++ b/Scripts/Gameflow.lua
@@ -6,11 +6,6 @@
 
 Flow.SetIntroImagePath("Screens\\main.jpg")
 
--- This image should be used for static title screen background (as in TR1-TR3).
--- For now it is not implemented.
-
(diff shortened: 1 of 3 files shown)


Lwmte:
This PR is dangerous. Since all references to this function are removed, all previous gameflow scripts will stop working. You should remove actual code for the function, its declaration in FlowHandler.h, and also unused variable `TitleScreenImagePath` (you forgot to), BUT - instead, you must provide a "deprecation" function which will just print a warning in the console that this function does not exist. Then you must assign this function to the name `SetTitleScreenImagePath`.

PR #1243, opened 2023-10-29
add View.GetAspectRatio() function

davidmarr:
add View.GetAspectRatio() lua function.
Corrected PlaySoundEffect() function.
Add Moveable:GetEndFrame() method to retrieve "End frame" of object's active animation


TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
```diff
@@ -265,6 +265,12 @@ void Moveable::Register(sol::table& parent)
 // @treturn int the current frame of the active animation
 	ScriptReserved_GetFrameNumber, &Moveable::GetFrameNumber,
 
+/// Retrieve End Frame.
+// This is the "End Frame" set in WADTool of the object's active animation.
+// @function Moveable:GetEndFrame
+// @treturn int End Frame of the active animation.	
+	ScriptReserved_GetEndFrame, & Moveable::GetEndFrame,
```
Kubsy, line 272: I am not sure if this is useful since users can just use GetFrameNumber and just type in the end frame themselves.

Kubsy (approved):

TombEngine/Scripting/Internal/TEN/Objects/Moveable/MoveableObject.cpp
Adngel: I think it would be a useful feature.

At this point, how would the build know which is the final frame of the animation? Right now the only way I know is to open Wadtool and see the value there.

For example here:

```lua
if (Lara:GetAnim() == LaraCustomAnimationStart + 1) then
	if (Lara:GetFrame() >= 9) then
```

Here, I'm checking to see if Lara is in the Lever Release animation and if she's at frame 9 or higher, then I continue. This works because in WadTool I see (and I know because I made it) that the animation is 10 frames.

However, there is a problem: if the animation changes (increasing or decreasing its duration), the builder would be forced to change all the constants in its code. In my case, now the last frame wouldn't be 9 but... something else.

But by using a function like GetFrameNumber, one can get the final frame of the animation without needing to figure out what it is.

Lwmte: Agreed

Kubsy:
pr looks good to me :) can be merged.

Lwmte (approved):

PR #1247, opened 2023-11-06
fix TR5 hydra power up flame and missile shockwave

Nemoel-Tomo:
just a small fix for the TR5 hydra
fixed shockwave light for hammer romangod


Stranger1992:
I tested both the `HYDRA` and `ROMAN_GOD1` .

I didn't spot anything with the Roman God and I really welcome the additional dynamic light on the ground attack.

The hydra's charge attack looks great (directly compared with the the original TRC). My only problem is that it took a VERY long time to kill the Hydra. I counted it took approximately 60 revolver shots to kill it, which is more powerful than any other enemy. I think that can be reduced but other than that it is good to go!

TombEngine/Game/effects/effects.cpp
```diff
@@ -1231,6 +1231,15 @@ void KillAllCurrentItems(short itemNumber)
 	// TODO: Reimplement this functionality.
 }
 
+void SpawnDynamicLight(const Vector3& pos, const Color& color, float falloff)
```
Lwmte, line 1234: Why this function is called `SpawnDynamicLight`, if we have a function called `TriggerDynamicLight` already? Make an overload.

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
```diff
@@ -27,6 +27,14 @@ namespace TEN::Entities::TR4
 	constexpr auto DEMIGOD1_WALK_RANGE					   = SQUARE(BLOCK(3));
 	constexpr auto DEMIGOD2_RADIAL_PROJECTILE_ATTACK_RANGE = SQUARE(BLOCK(5));
 	constexpr auto DEMIGOD3_RADIAL_PROJECTILE_ATTACK_RANGE = SQUARE(BLOCK(5));
+	constexpr auto DEMIGOD1_SHOCKWAVE_LIGHT_DURATION = 1.0f;
+
+	struct Demigod1Info
```
Lwmte, line 32: What is this?

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
```diff
@@ -27,6 +27,14 @@ namespace TEN::Entities::TR4
...
+	{
+		Vector3i LightPosition = Vector3i::Zero;
+	};
+
+	Demigod1Info Demigod1Data;
```
Lwmte, line 37: Why such complexity? You make a struct which keeps a vector3. You define a global variable which is this struct. Then you refer to this struct only to pass a vector3 from it to `SpawnDynamicLight`. Can't you just directly use this position to spawn it?

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
```diff
@@ -709,6 +730,15 @@ namespace TEN::Entities::TR4
 					TriggerShockwave((Pose*)&pos, 24, 88, 256, 128, 128, 128, 32, EulerAngles::Zero, 8, true, false, (int)ShockwaveStyle::Normal);
 					TriggerHammerSmoke(pos.x, pos.y + 128, pos.z, 8);
 
+					Demigod1Data.LightPosition = Vector3i(pos.x, pos.y, pos.z);
```
Lwmte, line 733: As I said above, you should not use a global.

Lwmte (changes requested):

TombEngine/Game/effects/effects.cpp
Sezzary: `SpawnDynamicLight()` is a replacement for `TriggerDynamicLight()` that uses the newer, consistently applied prefix convention decided upon for new or revised effect functions (`Spawn`).

Lwmte: I think you can do en masse rename in other PR then. Having 2 functions with same purpose but named differently is misleading.

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
Nemoel-Tomo: fixed

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
Nemoel-Tomo: fixed

TombEngine/Objects/TR4/Entity/tr4_demigod.cpp
Nemoel-Tomo: fixed

PR #1248, opened 2023-11-06
Convert debris to array

Lwmte:
Simplify debris code and possibly fix crash issues.

### DIFF
--- a/TombEngine/Game/control/control.cpp
+++ b/TombEngine/Game/control/control.cpp
@@ -305,7 +305,7 @@ GameStatus DoLevel(int levelIndex, bool loadGame)
 
 	// Initialize items, effects, lots, and cameras.
 	HairEffect.Initialize();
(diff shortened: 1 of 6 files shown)


Adngel:
Hi, I uploaded the level here:
https://drive.google.com/file/d/13Hpm2YcOEaFmWhNWr9gVUVjVyidb_ePB/view?usp=sharing

To replicate the bug, alternate killing some atlantean creatures with some centaurs. For example.

- I've got a crash after killing 2 Atlantean creatures and 1 centaur
- I've got a crash after killing 2 AC, 1 centaur, 1 AC and1 centaur
- I've got a crash after killing 1 AC, 1 centaur, 1 AC, 1 Giant Mutant

So is a bit random, but not very difficult to replicate if got alternating the kills. (I used the revolver, but I did not saw difference with the used weapon).

PR #1259, opened 2023-11-14
TR3- gun turret

Nemoel-Tomo:
TR3 Autogun

New moveable slot:     MOVEABLE#295
revised object is needed -> see attachment 

Once it is triggered, you can activate it by shooting at it. Or you can place a volume trigger and reduce its hitpoints to make it shoot at Lara.

[TR3GunTurret.zip](https://github.com/MontyTRC89/TombEngine/files/13357125/TR3GunTurret.zip)


TombEngine/Objects/TR3/Entity/gun_turret.cpp
```diff
@@ -0,0 +1,204 @@
...
+					auto bloodPos = GetJointPosition(&laraItem, Random::GenerateInt(0, NUM_LARA_MESHES - 1));
+					float bloodVel = Random::GenerateFloat(4.0f, 8.0f);
+					DoBloodSplat(bloodPos.x, bloodPos.y, bloodPos.z, bloodVel, Random::GenerateAngle(), laraItem.RoomNumber);
+					
+
+					pos = GetJointPosition(item, GunTurretLeftBite);
+					auto colorRed =		Random::GenerateFloat(0.75f, 1.0f) * UCHAR_MAX;
+					auto colorGreen =	Random::GenerateFloat(0.5f, 0.6f) * UCHAR_MAX;
+					auto colorBlue =	Random::GenerateFloat(0.0f, 0.25f) * UCHAR_MAX;
+					TriggerDynamicLight(pos.x, pos.y, pos.z, 2 * item.ItemFlags[0] + 8, colorRed, colorGreen, colorBlue);
```
Adngel, line 128: There is a new team convention where we are trying to avoid the use of the deprecated "GetRandomControl()" function. Also in case of these values that goes from 0 to 255, we are normalizing them between 0 to 1 and multiplying by UCHAR_MAX (255)

So a line like this:

GetRandomControl() & 0x3F) + 192

Is taking a random number [0, 63] + 192  it's to say a random value between [192, 255], what normalized are [0.75f , 1.0f]

That's why the new line got like this:

Random::GenerateFloat(0.75f, 1.0f) * UCHAR_MAX;


I changed other similar lines that I saw.

TombEngine/Objects/TR3/tr3_objects.cpp
```diff
@@ -332,6 +333,20 @@ static void StartEntity(ObjectInfo* obj)
 		obj->SetHitEffect();
 	}
 
+	obj = &Objects[ID_GUN_TURRET];
+	if (obj->loaded)
+	{
+		obj->Initialize = InitialiseGunTurret;
+		obj->control = GunTurretControl;
+		obj->collision = CreatureCollision;
```
Adngel, line 341: I know the turret use to be on ceiling and are unlikely for Lara to actually reaches to collide with her, but I though would be more natural to have a collision rutine instead of having her just traspassing it. That's why I added the standard one.

Adngel (approved):
Hi.

I've tested and reviewed the code.

About the code I made some notes and small changes here and there, but generally I see nothing wrong with it, for me is fine for it to merge.

The only issue that got me uncertain was the use of the AI_MODIFIER.

About the AI_MODIFIER I tested a level with a turret and that AI object in the same sector, I did not appreciate change on its behaviour (it starts idle, when Lara shoots at it, it starts to attack Lara). However, when I get cover, although the turret stop shooting, the shooting sound still happens, this bug only happes when AI_MODIFIER is applied, it doesn't happen with the normal gun turret.

However this issue with the AI_MODIFIER I don't think were something that really get influence the gameplay nor the main part of the code, so as I see it, it's fine to merge this branch.

Sezzary (approved):

PR #1275, opened 2023-11-30
keyholes: keep or loose key from inventory with ocb

Nemoel-Tomo:
I have included a new system for Keyholes. Now you can have one shot keyholes with a key trigger and multiple use keyholes with a switch trigger. But if you put a negative animnumber in the ocb of the keyhole you will loose the key from inventory and if you insert a positive animnumber into the ocb Lara will keep the key in the inventory.

One time keyhole with a key trigger: 
positive OCB -> play anim number and keep key in inventory.
negative OCB -> play anim number and loose key from inventory.

multiple keyhole with switch trigger: 
positive OCB -> play anim number and keep key in inventory.
negative OCB -> play anim number and loose key from inventory.

both: 
OCB = 0 play anim number 131 (LA_USE_KEY) and keeps the key in inventory

### DIFF
--- a/TombEngine/Objects/Generic/puzzles_keys.cpp
+++ b/TombEngine/Objects/Generic/puzzles_keys.cpp
@@ -445,6 +445,7 @@ void KeyHoleCollision(short itemNumber, ItemInfo* laraItem, CollisionInfo* coll)
 	if (triggerIndexPtr == nullptr)
 		return;
 
+	int animation;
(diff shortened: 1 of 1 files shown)


Lwmte:
I think it should be the opposite - positive or 0 OCB shoudl remove key from inventory, as it was by default.

Nemoel-Tomo:
> I think it should be the opposite - positive or 0 OCB shoudl remove key from inventory, as it was by default.

fixed

Adngel (approved):

Issue #834, opened 2022-11-11
Slope and savegame bug


Jakub768:
In the TEN level Angkor Adventure: https://trcustoms.org/levels/3474 if you land at the top of the slope, lara instead of sliding, she will leap. If you load the savegame while she is leaping, lara will spawn in a void

https://user-images.githubusercontent.com/80340234/201423208-4c0e6222-bd17-4f2d-8709-91a70e3281a0.mp4

here's a savegame where you can repro the issue in the level
[savegame.zip](https://github.com/MontyTRC89/TombEngine/files/9992998/savegame.zip)

PR #1277, opened 2023-12-02
Bridge refactor

Sezzary:
Adds `BridgeObject` `ItemData` variant and moves bridge routines from `ObjectInfo` into it.

For code review:
- Check if all bridge objects have been accounted for in the object ID collection.
- Maybe there's a better way to do this? To load bridges before everything else, I substitute a hack for a hack.

To test:
- Climbable pushables placed on bridges on level load.
- Non-bridge moveables placed on bridges on level load.
- Ensure bridges still work the same as before.


Kubsy:
So after some tests I can say that bridges almost are correct!

I have tested for these 2 issues and couldn't reproduce them which I guess they got fixed now:
image

and the saving on bridges is fixed too - lara doesn't teleport below it anymore. https://github.com/MontyTRC89/TombEngine/issues/834

However I did notice some stuff with pushables on bridges:

1) I could not pull the pushable from the bridge side:
https://github.com/MontyTRC89/TombEngine/assets/80340234/b217b5e6-f4f4-4cd8-af19-cd7300a46d1b

2) (these 2 issues seem to appear when lara is on a last bridge where lara cannot pull anymore) 
- Lara gets teleported.
- Lara gets stuck inside the pushable

https://github.com/MontyTRC89/TombEngine/assets/80340234/21448e13-3a1d-4c32-8842-8d97cbe77022

Uploading Desktop 2023.12.05 - 12.05.00.03.mp4…

Here's the level to have a play around with

Kubsy:
but apart from that, I see no issues with bridges.

Lwmte (approved):

PR #1278, opened 2023-12-03
Management reasons end of level as separate events

davidmarr:
Management reasons end of level as separate events

### DIFF
--- a/TombEngine/Game/control/control.cpp
+++ b/TombEngine/Game/control/control.cpp
@@ -481,16 +481,16 @@ void DeInitializeScripting(int levelIndex, GameStatus reason)
 	switch (reason)
 	{
 	case GameStatus::LoadGame:
-		HandleAllGlobalEvents(EventType::BeforeLoad, (Activator)LaraItem->Index);
+		HandleAllGlobalEvents(EventType::LoadGame, (Activator)LaraItem->Index);
 		break;
 	case GameStatus::ExitToTitle:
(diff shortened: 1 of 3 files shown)


TombEngine/Game/control/control.cpp
```diff
@@ -478,7 +478,23 @@ void InitializeScripting(int levelIndex, bool loadGame)
...
+		HandleAllGlobalEvents(EventType::BeforeLoad, (Activator)LaraItem->Index);
+		break;
+	case GameStatus::ExitToTitle:
+		HandleAllGlobalEvents(EventType::ExitToTitle, (Activator)LaraItem->Index);
+		break;
+	case GameStatus::LaraDead:
+		HandleAllGlobalEvents(EventType::Death, (Activator)LaraItem->Index);
+		break;
+	case GameStatus::LevelComplete:
+		HandleAllGlobalEvents(EventType::End, (Activator)LaraItem->Index);
```
Lwmte, line 493: Can we keep same names for all events, in correlation with `GameStatus`? E.g. if GameStatus is `LoadGame`, can event be also called `EventType:: LoadGame`, and so on for all other events?

TombEngine/Game/control/event.h
```diff
@@ -41,6 +41,9 @@ namespace TEN::Control::Volumes
 		Save,
 		Start,
 		End,
+		BeforeLoad,
```
Lwmte, line 44: I think you need to additionally extend exposed Lua enumeration (similar to what I did in my last 2 commits in main PR).

PR #1284, opened 2023-12-11
New animation sound conditions

Sezzary:
Revises animation sound effect conditions. Will require asset updates for water creatures later, but for now their sounds are handled with a special exception.

Old sound conditions:
- `Always`
- `Land`
- `Water`

New sound conditions:
- `Always`
- `DryLand`
- `WetLand`
- `Swamp`
- `Underwater`

To test:
- `Underwater` sounds should only play when the camera is submerged.
- `Swamp` sounds shoudl play only in swamps.
- `WetLand` sounds should play in shallow water as before. Check on sloped surfaces.


Lwmte:
I propose to leave `Land` as `Land`, and rename `WetLand` to `ShallowWater`.

TombEngine/Sound/sound.cpp
```diff
@@ -146,44 +149,46 @@ bool LoadSample(char* pointer, int compSize, int uncompSize, int index)
...
-	if (condition != SoundEnvironment::Always)
+	// Test if sound effect environment matches camera environment.
+	if (soundEnv != SoundEnvironment::Always)
 	{
-		// Get current camera room's environment
-		auto cameraCondition = TestEnvironment(ENV_FLAG_WATER, Camera.pos.RoomNumber) ? SoundEnvironment::Water : SoundEnvironment::Land;
-
-		// Don't play effect if effect's environment isn't the same as camera position's environment
-		if (condition != cameraCondition)
+		bool isCameraUnderwater = TestEnvironment(ENV_FLAG_WATER, Camera.pos.RoomNumber);
```
Lwmte, line 167: Shouldn't you add similar condition for swamp environment type?

TombEngine/Sound/sound.cpp
```diff
@@ -192,77 +197,85 @@ bool SoundEffect(int effectID, Pose* position, SoundEnvironment condition, float
...
-		pitch += ((static_cast<float>(GetRandomControl()) / static_cast<float>(RAND_MAX)) - 0.5f) * SOUND_MAX_PITCH_CHANGE * 2.0f;
+	if ((sample.Flags & SOUND_FLAG_RND_PITCH))
+		pitch += Random::GenerateFloat(-0.5f, 0.5f) * (SOUND_MAX_PITCH_CHANGE * 2);
 
-	// Calculate sound radius and distance to sound
-	float radius = (float)(sampleInfo->Radius) * BLOCK(1);
-	float distance = Sound_DistanceToListener(position);
+	// Calculate sound radius and distance to sound.
+	float radius = BLOCK(sample.Radius);
+	float dist = Sound_DistanceToListener(pose);
```
Lwmte, line 222: Usually you make variable names more verbose and descriptive, but here you decided to do the opposite and convert `distance` to `dist`. Why?

TombEngine/Sound/sound.cpp
```diff
@@ -1056,26 +1070,27 @@ int GetShatterSound(int shatterID)
 		return SFX_TR4_SMASH_ROCK;
 }
 
+// TODO: Revise flags.
```
Lwmte, line 1073: Which flags you want to revise?

TombEngine/Sound/sound.cpp
```diff
@@ -1056,26 +1070,27 @@ int GetShatterSound(int shatterID)
...
-		int group = sound.Flags & 0x1FFF;
+		int group = soundSource.Flags & 0x1FFF;
 
 		if (group >= MAX_FLIPMAP)
 			continue;
 
-		if (!FlipStats[group] && (sound.Flags & PLAY_FLIP_ROOM))
-			continue;
-		else if (FlipStats[group] && (sound.Flags & PLAY_BASE_ROOM))
+		bool flipStat = FlipStats[group];
```
Lwmte, line 1087: I am not sure why you need a temp variable if array is accessed only twice and is already readable (it's not like we have an array deeply hidden in some class like `GameEngine.FlipmapController->FilpStats[flipInfo.group]`).

TombEngine/Sound/sound.cpp
Sezzary: The `Underwater` sound condition is unique in that sounds will only play only when the camera is submerged. The other conditions are "above water".

TombEngine/Sound/sound.cpp
Sezzary: I use the full long word in more formal contexts such as a class name, function name, or member variable, but for local variables I go for brevity. Same with pos/position, orient/orientation, rot/rotation, etc.

TombEngine/Sound/sound.cpp
Sezzary: I think this was a note for myself to make a flag enum.

TombEngine/Sound/sound.cpp
Sezzary: Sure!

TombEngine/Sound/sound.cpp
Lwmte: Yes, that's what I mean. If object is submerged into the swamp, shouldn't sounds also be canceler? For example, there is a funny bug when Lara dies in the swamp, you can hear her grunt.

TombEngine/Sound/sound.cpp
Lwmte: But I still don't understand why this choice.

TombEngine/Sound/sound.cpp
Sezzary: `Underwater` is unique in that sense. `Swamp` works more like `ShallowWater`.

TombEngine/Sound/sound.cpp
Sezzary: Consistency.

Lwmte: I understand that you try to maintain consistency with your own code in other places of the engine. However, it is not indicated in coding conventions, and this is the first time I am hearing this from you and seeing this.

TombEngine/Sound/sound.cpp
Lwmte: I get it then. Would playback condition be checked against `Underwater` case if sound is emitted within the swamp room then?

Adngel:
I've tried but I'm getting compiling errors.

In Jeep is DryLand, so now it should "Land"?

In animation.cpp line case AnimCommandType::Flipeffect:, the compiler says the initialization of frameNumber is omitted in this case. Maybe a version conflict affected this PR somehow?

PR #1327, opened 2024-02-04
Revised spiky wall

Nemoel-Tomo:
To stop the wall, just make a volume trigger and set the OCB of the wall to 0. To make it move again, create a volume trigger with OCB and the desired speed value.

### DIFF
--- a/TombEngine/Objects/TR4/Trap/SpikyWall.cpp
+++ b/TombEngine/Objects/TR4/Trap/SpikyWall.cpp
@@ -18,7 +18,7 @@ using namespace TEN::Math;
 
 // NOTES:
 // ItemFlags[0]: forward velocity.
-// ItemFlags[1]: 
+// ItemFlags[1]: enable collision with other spiky walls.
 
 namespace TEN::Entities::Traps
 {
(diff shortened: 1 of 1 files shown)


TombEngine/Objects/TR4/Trap/SpikyWall.cpp
```diff
@@ -23,37 +29,88 @@ namespace TEN::Entities::Traps
...
+					abs(pointColl1.Coordinates.z - mesh.pos.Position.z) < BLOCK(1) &&
+					StaticObjects[mesh.staticNumber].shatterType != ShatterType::None)
+				{					
+					if (mesh.HitPoints != 0)
+						continue;
+
+					mesh.HitPoints -= 1;
+					ShatterObject(nullptr, &mesh, -64, LaraItem->RoomNumber, 0);
+					SoundEffect(SFX_TR4_SMASH_ROCK, &item.Pose);
+					TestTriggers(item.Pose.Position.x, item.Pose.Position.y, item.Pose.Position.z, item.RoomNumber, true);
```
Adngel, line 99: In the future, we should make a function in the shatters objects that includes this "shattering" routine, so when any object wants to break them, (Lara shooting, rollingballs, spikewalls, enemies, etc) the objects always do it by their same routine. 

I'm concerned specially for the sound constant value, if at some point we neee to change it so each shatter have uses own sound, (rock, glass, wood, etc), having that in a single routine should be easier.

But as I see it, that's a task different of the purpose of this branch, so is is fine as it's, we better cover that in a different branch in the future, (as it may need to edit other objects as well)

Adngel (approved):
I saw not issue on the code, nor the gameplay test, so I see fine to implement this branch in the develop one.

PR #1330, opened 2024-02-12
Ldoc portable compiler

Lwmte:


Stranger1992 (changes requested):
The build event inside Visual Studio works well and applies the changes correctly. However: all HTML files are added to the github changes list. Can this be avoided?

PR #1332, opened 2024-02-24
Fish Emitter

Nemoel-Tomo:
OCB 0 = does nothing
OCB 1 = mesh01
OCB 2 = mesh02
OCB 3 = mesh03   and so on...

negative numbers means lethal

OCB -1= mesh 01 (lethal)
OCB -2= mesh 02 (lethal)  and so on...

HitPoints = fish number. (If you use volume triggers to change fish number please choose HP **"set"** to new value.

the fish use the meshes from the FISH_EMITTER object. mesh00 should be a common nullmesh

You can make the fish patrol between AI_FOLLOW objects:
You can use any number of Ai objects for a single path.
A path requires at least 3 AI_Follow objects. The one with the first OCB has to be on the start square of the fishemitter, and then the fish swim back and forth between the next other two. After the AI_FOLLOW with the last OCB value, the fish automatically swims back to the AI_FOLLOW with the first valid OCB within the path. (the first after the starting AI_Follow Object)
 If you then want to make a another path for another school of fish, you simply leave out an OCB number between the last ocb of the previous path and the start number of the new path. This way the engine knows which numbers belong to which swarm. the first AI_FOLLOW (the one that's on the same square as the fishemitter does not count for patrolling but also needs its OCB). If you do not place an AI_Follow object on the fish emitter square, the fish will follow a random target within a solid spheroid within the boundaries of the water room.
 
Here is an example: Between the last ocb (4) and start of the new path (OCB 6) has to be a gap (no object with OCB 5) so that the engine knows there are 2 different paths.
image


Nemoel-Tomo:
[fishemitter.zip](https://github.com/MontyTRC89/TombEngine/files/14396765/fishemitter.zip)

Nemoel-Tomo:
TODO: Save and load fish needs to be implemented before approving. (thx Sezz)

TombEngine/Objects/TR3/Entity/FishSwarm.cpp
```diff
@@ -0,0 +1,410 @@
...
+		ai.distance = SQUARE(dx) + SQUARE(dz);
+
+		item.Animation.Velocity.z = FISH_VELOCITY_MAX;
+
+		auto& playerRoom = g_Level.Rooms[LaraItem->RoomNumber];
+
+		auto corpsePos = std::optional<Vector3>();
+
+		// Check if corpse is near.
+		if (!corpsePos.has_value())
```
Lwmte, line 139: I don't fully understand why there is a `has_value()` check for `corpsePos`, which was just declared one line above.

TombEngine/Objects/TR3/Entity/FishSwarm.cpp
```diff
@@ -0,0 +1,410 @@
...
+		{
+			float closestDist = INFINITY;
+			for (auto& targetItem : g_Level.Items)
```
Lwmte, line 142: Cycling through all items in a level every frame (and for every fish swarm) may become extremely slow. Can we consider narrowing search, for example, by searching only in neighbor rooms of a fish swarm? Additionally, this search may be done not every single frame, but e.g. every second or fourth frame (this can be achieved by using bitwise operation against `GlobalCounter` variable.

TombEngine/Objects/TR3/Entity/FishSwarm.cpp
```diff
@@ -0,0 +1,410 @@
...
+			auto orientTo = Geometry::GetOrientToPoint(fish.Position, desiredPos.ToVector3());
+			fish.Orientation.Lerp(orientTo, 0.1f);
+
+			for (const auto& otherFish : FishSwarm)
+			{
+				if (&fish == &otherFish)
+					continue;
+
+				float distToOtherFish = Vector3i::Distance(fish.Position, otherFish.Position);
+				float distToPlayer = Vector3i::Distance(fish.Position, LaraItem->Pose.Position);
```
Lwmte, line 310: Probably a note for the future, but if we somehow manage to detect corpses to attack by fish swarm, maybe we can also detect any creature in the water, not only Lara? Then you can arrange different scenarios, such as shark or crocodile attacked by pirahnas, and so on.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -610,14 +611,129 @@ namespace TEN::Renderer
...
+		using namespace TEN::Entities::Creatures::TR3;
+
+		if (!Objects[ID_FISH_EMITTER].loaded)
+			return;
+
+		if (rendererPass == RendererPass::CollectTransparentFaces)
+		{
+			for (const auto& fish : FishSwarm)
+			{
+				if (fish.Life > 0.0f)
```
Lwmte, line 625: Can we invert this condition to `if (fish.Life <= 0.0f) continue;` to avoid excessive tabulation?

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -610,14 +611,129 @@ namespace TEN::Renderer
...
+				{
+					auto& mesh = *GetMesh(Objects[ID_FISH_EMITTER].meshIndex + fish.MeshIndex);
+
+					for (auto& bucket : mesh.Buckets)
+					{
+						if (IsSortedBlendMode(bucket.BlendMode))
```
Lwmte, line 631: Same here, invert condition to avoid excessive tabulation.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -610,14 +611,129 @@ namespace TEN::Renderer
...
+				_context->IASetVertexBuffers(0, 1, _moveablesVertexBuffer.Buffer.GetAddressOf(), &stride, &offset);
+				_context->IASetIndexBuffer(_moveablesIndexBuffer.Buffer.Get(), DXGI_FORMAT_R32_UINT, 0);
+
+				const auto& moveableObj = *_moveableObjects[ID_FISH_EMITTER];
+
+				_stStatic.LightMode = (int)moveableObj.ObjectMeshes[0]->LightMode;
+
+				for (const auto& fish : FishSwarm)
+				{
+					if (fish.Life > 0.0f)
```
Lwmte, line 693: Same as above.

Frjttr:
Tests were fine on my side. Didn't test again the OCB 0 for the AI_Follow obj

TombEngine/Renderer/RendererDraw.cpp
Nemoel-Tomo: fixed

TombEngine/Renderer/RendererDraw.cpp
Nemoel-Tomo: fixed

Stranger1992 (changes requested):
Current issues (1st April 2024)

* Fish Emitter object will crash game if all 5 code-bits are set to on (ie: to trigger on start up)
* Blood - in - water effect is missing when swarm bite 

https://github.com/MontyTRC89/TombEngine/assets/84292688/ffda9263-c000-4bf0-ae17-73433fc50f45

* If near a wall / room boundary: the fish mesh will change. The below video has this problem when the fish reach an AI_FOLLOW with a OCB set to 3

https://github.com/MontyTRC89/TombEngine/assets/84292688/c49e9835-d079-475d-a006-1c16362522f1

* Fish (with OCB -1) are ignoring CORPSE object. user will need to first trigger the CORPSE object and the fish will only bite it if Lara is in the water:

https://github.com/MontyTRC89/TombEngine/assets/84292688/3e8b052d-5fde-4586-a939-6edbe038fdb0

* Trying to save game results in a crash 
* Trying to load game results in a crash

Stranger1992:
> Current issues (1st April 2024)
> 
> * Fish Emitter object will crash game if all 5 code-bits are set to on (ie: to trigger on start up)
> * Blood - in - water effect is missing when swarm bite
> 
>  2024-04-01_02-03-44.mp4 
> * If near a wall / room boundary: the fish mesh will change. The below video has this problem when the fish reach an AI_FOLLOW with a OCB set to 3
> 
>  2024-04-01_02-07-51.mp4 
> * Fish (with OCB -1) are ignoring CORPSE object. user will need to first trigger the CORPSE object and the fish will only bite it if Lara is in the water:
> 
>  2024-04-01_02-14-17.mp4 
> * Trying to save game results in a crash
> * Trying to load game results in a crash

After the latest commits, based on my previous testing:

* All above fixed. I was still having where save/load would crash the game so I deleted my SAVE folder and tried again and that worked. Fish positions were remembered correctly.
*

PR #1335, opened 2024-03-07
EmberEmitter

Nemoel-Tomo:
Object ID: 
ID_EMBER_EMITTER 
Object ID: 1057

OCB 0 = Lava spitting effect from TR1 and TR2 in its original size (about 1,6 Block height) The effect is harmful to Lara. To change the size just insert positive numbers into the ocb: 11 will be just about 1.1 Blocks, 20 will be about 2.0 blocks.

Choose the color of the effect via OCB Tint
Only black (r,g,b = 0) will have a subtractive blending mode and appears black 

negative OCB: Lava spitting effect from TR5 (Ireland Labyrinth level). The effect is harmful to Lara. The OCB value covers the area of the effect. (In TR5 a value of 31 is used)


TombEngine/Objects/Effects/EmberEmitter.cpp
```diff
@@ -0,0 +1,157 @@
...
+		auto& item = g_Level.Items[itemNumber];
+
+		if (!TriggerActive(&item))
+			return;
+
+		unsigned char r = 0;
+		unsigned char g = 0;
+		unsigned char b = 0;
+
+		if (item.Model.Color.x == item.Model.Color.y &&
```
Lwmte, line 23: I don't fully understand this if-then condition. Does it mean that if object tint is grayscale, no brightness variation in color will occur?

Nemoel-Tomo: Yes, with the gray color, the more the color goes into black, the more it would look like rainbow confetti because of the blend mode. Since I want to keep the rainbow confetti as an Easter egg, and because I also want to keep the color gray, I have excluded the brightness variation with a gray value where r, g and b are the same.

Lwmte: What if you just variate all values (RGB) using same random number? Then they will remain grayscale but still will variate in brightness. In fact, I think similar approach must be used for any colour, this will prevent color drift (e.g. red particles become slightly magenta and so on).

Nemoel-Tomo: ah ok, I will try that

Nemoel-Tomo: yes, this fixed the issue

TombEngine/Objects/Effects/EmberEmitter.cpp
```diff
@@ -0,0 +1,151 @@
...
+	void ControlEmberEmitter(short itemNumber)
+	{
+		auto& item = g_Level.Items[itemNumber];
+
+		if (!TriggerActive(&item))
+			return;
+
+		unsigned char r = 0;
+		unsigned char g = 0;
+		unsigned char b = 0;
```
Jakub768, line 21: you can put 3 vars into one like:

unsigned char r = 0, g = 0, b = 0

PR #1341, opened 2024-03-29
Load spot cameras properly

Sezzary:
Minor revision to avoid loading spot cameras as a chunk. Otherwise, modifying the `SPOTCAM` struct in any way would break everything.

To test, ensure spot cameras still work as before.

### DIFF
--- a/TombEngine/Specific/level.cpp
+++ b/TombEngine/Specific/level.cpp
@@ -85,42 +85,42 @@ std::vector<int> StaticObjectsIds;
 std::vector<int> SpriteSequencesIds;
 LEVEL g_Level;
 
-unsigned char ReadUInt8()
+static unsigned char ReadUInt8()
 {
 	unsigned char value = *(unsigned char*)LevelDataPtr;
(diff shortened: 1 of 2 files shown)


TombEngine/Specific/level.cpp
```diff
@@ -85,50 +85,50 @@ std::vector<int> StaticObjectsIds;
 std::vector<int> SpriteSequencesIds;
 LEVEL g_Level;
 
-unsigned char ReadUInt8()
+static unsigned char ReadUInt8()
```
Lwmte, line 88: Why do you introduce `static` keyword for global functions, which are already implcitly static?

Sezzary: This goes back to a discussion we had last year about best practice in this regard. It limits the visibility of functions to the specific file in which they are used (reducing the risk of cryptic naming conflicts upon compilation) and helps encapsulate the implementation. IMO I find the best benefit is readability, since it visually indicates when a function is only intended for use within a specific file.

PR #1346, opened 2024-04-11
Raptor upgrade

TokyoSU:
- This include the ability to jump up and down 2/3/4 click height, jump pit of 1 or 2 block.
- Fixed the running angle being 2 where it should be 4, more reactive when moving now :)


GeckoKid-JW:
Currently when the raptors die when vaulting, to prevent them dying in mid-air, they would teleport to the ground level and die. However an unexpected consequence is that the raptors would play their death animations multiple times occasionally.

I propose making their death animation only play when they are not vaulting, sort of like the t-rex and bear where it only plays the death animation after they have finished what they were doing (like roar, standing, attacking etc).

Stranger1992:
Hi @TokyoSU can you let me know if this PR is ready to test or it is in development still?

TokyoSU:
> Hi @TokyoSU can you let me know if this PR is ready to test or it is in development still?

This need a ocb flags to allow/remove jump ability and the death animation need to be after the jump finish and not in the middle.

Stranger1992:
> > Hi @TokyoSU can you let me know if this PR is ready to test or it is in development still?
> 
> This need a ocb flags to allow/remove jump ability and the death animation need to be after the jump finish and not in the middle.

Okay. Can you please merge the develop branch into this? I had trouble because the differences were so great.

PR #1359, opened 2024-04-27
Tomo TR-LastSaga-5 Waterfall_Emitter

Nemoel-Tomo:
Please test. 
Object_ID= 1088
Width = OCB. Width of the waterfall is calculated with CLICK. If you want to have a BLOCK then you have to multiply the OCB value by 4 and enter it in the OCB.
Particles should stop at the floor or on water/swamp surfaces.
Color = Object Tint
Sprites: by GeckoKid, see attachment:

now they are even more customizeable 

        //Customize x and z vel:
        item.ItemFlags[0] = 100;

        //Customize Waterfall Sprite scale
        item.ItemFlags[1] = 3;

        //Customize density
        item.ItemFlags[2] = 120;

        //Customize Waterfallmist Sprite scale
        item.ItemFlags[3] = 3;

        //Customize Sound (on or off)
        item.ItemFlags[4] = 1 no sound
        item.ItemFlags[4] = 0 sound

New Object for Sprites (20.09.2024)

[WaterfallEmitterSprites_3.zip](https://github.com/user-attachments/files/17072761/WaterfallEmitterSprites_3.zip)


Adngel (commented):
Hi, I've been reviewing this branch, and I think it look great overalll. I just have one suggestion regarding the "while (true)" loop in the WaterfallControl function.

Although I'm understanding it works, using while (true) can sometimes make the code harder to read and therefore, maintain. It might be clearer to use a loop with an explicit exit condition. This way, anyone reading the code in the future can understand the loop's purpose and exit conditions, more easily.

Or even a "do-while" loop could work well here and be closer to your original idea of do the action 1 time, and repeat if needed.

For example:
```c++
int currentStep = 0;
int offset;

do
{
    offset = (step * currentStep) + Random::GenerateInt(-32, 32);

    for (int sign = -1; sign <= 1; sign += 2)
    {
        if (Random::GenerateInt(0, 100) > std::clamp((width / 100) * 3, 30, 80))
        {
            // Instantiate the particles code.
        }

        if (sign == 1)
        {
            currentStep++;
            if (currentStep == 1)
                break;
        }
    }
} while (offset <= width);
```
of if prefer keep the while

```c++
int currentStep = 0;
int offset = 0;

while (offset <= width)
{
    offset = (step * currentStep) + Random::GenerateInt(-32, 32);

    for (int sign = -1; sign <= 1; sign += 2)
    {
        if (Random::GenerateInt(0, 100) > std::clamp((width / 100) * 3, 30, 80))
        {
            // Instantiate the particles code.
        }

        if (sign == 1)
        {
            currentStep++;
            if (currentStep == 1)
                break;
        }
    }
}
```

I think any of these two versions are more usual practices and could help the readability of the code.

Apart of that, I need to test the object before give my green light but I didn't saw any other issue in the code.

Stranger1992:
PR converted to draft whilst it is still being worked on. Once testing can begin with it: please change it.

TrainWrack:
Adding feedback here as well.

I tested the branch and currently I see no difference between OCB 1-9. It shows the same size of the waterfall. 

Also, when there are too many waterfalls active the splash effect starts appearing everywhere on the floor. Could be cause of particle limit?

I tried with just three active waterfalls OCB 20 and 30 and even then particles appeared.

https://www.youtube.com/watch?v=1Lt4Y-rib3w

TombEngine/Game/effects/effects.cpp
```diff
@@ -35,8 +38,9 @@ using namespace TEN::Effects::Explosion;
 using namespace TEN::Effects::Items;
 using namespace TEN::Effects::Ripple;
 using namespace TEN::Effects::Spark;
+using namespace TEN::Effects::WaterfallEmitter;
 using namespace TEN::Math;
-using namespace TEN::Math::Random;
+using namespace TEN::Math::Random; // TODO: No.
```
Lwmte, line 43: I recommend to avoid such meaningless comments.

TombEngine/Game/effects/effects.h
```diff
@@ -162,6 +176,8 @@ struct Particle
 	byte PrevG; 
 	byte PrevB;
 	byte PrevScalar;
+	int itemNumber = 0;
```
Lwmte, line 179: You can reuse `fxObj` flag instead of creating new field here.

TombEngine/Game/savegame.cpp
```diff
@@ -1151,7 +1151,7 @@ const std::vector<byte> SaveGame::Build()
 		particleInfo.add_b(particle->b);
 		particleInfo.add_col_fade_speed(particle->colFadeSpeed);
 		particleInfo.add_d_b(particle->dB);
-		particleInfo.add_sprite_index(particle->spriteIndex);
+		particleInfo.add_sprite_index(particle->SpriteSeqID); // TODO: Save SpriteID.
```
Lwmte, line 1154: It shouldn't be a todo, it should be added in this PR.
Additionally, `targetPos` field also should be saved in a savegame, otherwise your particle will fall through the floor and won't be disabled in time.

TombEngine/Objects/TR5/Emitter/Waterfall.cpp
```diff
@@ -0,0 +1,216 @@
...
+        part.PrevX = pos.x;
+        part.PrevY = pos.y;
+        part.PrevZ = pos.z;
+		part.x = pos.x;
+		part.y = Random::GenerateInt(-16, 0) + pos.y;
+		part.z = pos.z;
+        part.roomNumber = roomNumber;
+
+		// TODO: Generate normal color representation, then convert to legacy.
+		char colorVariation = (Random::GenerateInt(-8, 8)); // TODO: Why char?
```
Lwmte, line 182: since you cast to `(int)` down the line, please use `int` here as well :)

TombEngine/Objects/game_object_ids.h
```diff
@@ -998,7 +999,8 @@ enum GAME_OBJECT_ID : short
 	ID_AIR_BAR_TEXTURE,
 	ID_DASH_BAR_TEXTURE,
 	ID_SFX_BAR_TEXTURE,
-	// NOTE: 1378 - 1379 reserved for blood effects. -- Sezz 2023.05.29
+	ID_WATERFALL,
```
Lwmte, line 1002: We have pending blood effect pull request from Sezz, probably it's better to use another ID.

Lwmte (changes requested):

TombEngine/Objects/TR5/Emitter/Waterfall.cpp
```diff
@@ -0,0 +1,220 @@
...
+            part.dSize = item.TriggerFlags < 10 ? Random::GenerateFloat(40.0f, 51.0f) : Random::GenerateFloat(98.0f, 174.0f);
+            part.flags = SP_SCALE | SP_DEF | SP_ROTATE;
+        }
+    }
+
+	void SpawnWaterfallMist(const Vector4& pos, int roomNumber, float scalar, float size, const Color& color)
+	{
+		auto& part = *GetFreeParticle();
+
+		auto colorOffset = Color(40.0f, 40.0f, 40.0f); // make constant. Color is Vector 4. 4th value should not be changed.
```
Lwmte, line 166: `Color` class is used in the wrong way here. `Color` class expects normalized values to be used (from 0.0f to 1.0f). Please convert `color` argument to either `Vector3i` or individual `byte` values, and also don't use `Color` class in similar contexts down the line.

Lwmte (changes requested):

TombEngine/Specific/savegame/schema/ten_savegame.fbs
```diff
@@ -368,6 +368,9 @@ table ParticleInfo {
 	fx_obj: int32;
 	room_number: int32;
 	node_number: int32;
+	targetPos: Vector4;
+	SpriteSeqID: GAME_OBJECT_ID;
```
Lwmte, line 372: Flatbuffers specification requires all fields to be in a lowercase, and do not use CamelCase. So instead of `SpriteSeqID`, it should be `sprite_seq_id`, and instead of `targetPos`, it should be `target_pos`.

TombEngine/Game/effects/effects.cpp
```diff
@@ -201,69 +208,87 @@ void UpdateSparks()
...
+					ParticleDynamics[spark.dynamic].On = false;
 
-				spark->on = false;
+				spark.on = false;
+				continue;
+			}
+
+			if (spark.SpriteSeqID == ID_WATERFALL &&
+				spark.y >= spark.targetPos.y && 
+				spark.targetPos.w == spark.fxObj)
```
Lwmte, line 230: Why item number is packed into `w` component of `targetPos`, and why it is compared here with `fxObj` field of the very same spark? Can spark ownership change in runtime, or it's just duplicated fields?

TombEngine/Objects/Generic/Object/burning_torch.cpp
```diff
@@ -72,7 +72,7 @@ namespace TEN::Entities::Generic
 		spark->dSize = spark->size / 8;
 
 		int spriteOffset = GlobalCounter % Objects[ID_FIRE_SPRITES].nmeshes;
-		spark->spriteIndex = Objects[ID_FIRE_SPRITES].meshIndex + spriteOffset;
```
Lwmte, line 75: Are we missing `SpriteSeqID` assignment here?

TombEngine/Scripting/Internal/TEN/Objects/ObjectIDs.h
```diff
@@ -1924,7 +1924,7 @@ static const std::unordered_map<std::string, GAME_OBJECT_ID> kObjIDs {
 	{ "SMOKE_EMITTER", ID_SMOKE_EMITTER },
 	{ "EARTHQUAKE", ID_EARTHQUAKE },
 	{ "BUBBLES", ID_BUBBLES },
-	{ "WATERFALLMIST", ID_WATERFALLMIST },
+	{ "WATERFALL_MIST", ID_WATERFALL_MIST },
```
Lwmte, line 1927: Object ID changes should be synchronous with changes in TE repo, otherwise nodes which operate on slot IDs may break.

TombEngine/Renderer/RendererDrawEffect.cpp
```diff
@@ -497,10 +497,10 @@ namespace TEN::Renderer
 				}
 
 				// Disallow sprites out of bounds.
-				int spriteIndex = std::clamp((int)particle.spriteIndex, 0, (int)_sprites.size());
+				int spriteIndex = std::clamp((int)particle.SpriteID, 0, (int)_sprites.size());
```
Lwmte, line 500: I am afraid this clamp will not work properly anymore, because `spriteIndex` was converted to `SpriteID` and `SpriteSeqID`.

Lwmte (approved):

Stranger1992 (approved):
Final testing.

_All below tested with OCB 4_
# Do they trigger correctly in game? YES

image

image

# ItemFlag [0]:

itemflag[0] = 400:

image

itemflag[0] = 2:

image

# itemFlag[1]

itemflag[1] = 12:
image

itemflag[1] = 1:

image

# itemflag[2]

itemflag[2] = 12 

image

itemflag[2] = 1 

image

# itemflag[3]

itemflag[3] = 12

image

_NB: A little hard to see but the sprites were HUGE_

itemflag[3] = 0

image

_Sprites have disappeared entirely, as expected! _

# itemflag[4]

Sound was turned off if = 1

# Particle Test

10 Flame emitters + 1 waterfall emitter:

image

image

* No sprites missing. 
* All flame emitters working.

10 Flame emitters + 5 waterfall emitter + 5 lit flares creating bubbles in water:

image

# Conclusion

AMAZING work. Loved this new effect from its inception and seeing far it has bee developed. Really great work Nemoel!!!! Thank you for your contribution! :D :D :D :D 



image

PR #1362, opened 2024-05-07
60 FPS, starfield, lens flare

Jakub768:
By MontyTRC

This PR introduces: 

* Variable framerate (allowing a framerate more then 29.97 FPS and provides smoother graphics
* Lensflare
* Starfield

Please test and communicate any issues on #1365


Stranger1992:
@MontyTRC89 Hi. As this PR is still in development from what I can see. Until it gets to a state that you feel is good to commence testing: I have made it into a draft.

TombEngine/Game/effects/weather.cpp
```diff
@@ -9,44 +9,24 @@
...
+#include "Specific/level.h"
 
 using namespace TEN::Collision::Point;
 using namespace TEN::Effects::Ripple;
-using namespace TEN::Math::Random;
+using namespace TEN::Math;
 
 namespace TEN::Effects::Environment 
 {
-	constexpr auto WEATHER_PARTICLES_SPAWN_DENSITY = 32;
```
Lwmte, line 23: Why these constants were moved to .h file, if their scope is only this particular module? There is no need to expose them on global scale.

TombEngine/Game/Hud/PickupSummary.h
```diff
@@ -25,8 +25,21 @@ namespace TEN::Hud
...
 
+		Vector2		PrevPosition	= Vector2::Zero;
+		EulerAngles PrevOrientation = EulerAngles::Identity;
+		float		PrevScale		= 0.0f;
+		float		PrevOpacity		= 0.0f;
+
 		bool IsOffscreen() const;
 		void Update(bool isHead);
+
+		void StoreInterpolationData()
```
Lwmte, line 36: Regarding this, and further additions of `StoreInterpolationData`: this is a case for an abstract class. We should think about inheriting all structs/classes which implement `StoreInterpolationData` from the same base.

MontyTRC89: Agreed

TombEngine/Game/effects/weather.cpp
MontyTRC89: It was for having all constants in the same place :-)

PR #1392, opened 2024-08-08
Seal Mutant Port

TokyoSU:
## todo

None

- Port the seal mutant from TR3 to TEN.
- Improved by adding OCB 1 so it enable the TRAP when he goes up, throw gas then die.

Required wad assets:
- Need to fix the sounds.
[SealMutant.zip](https://github.com/user-attachments/files/16550720/SealMutant.zip)


Stranger1992 (approved):
Play testing in-game has thrown no errors. Not a very complicated enemy but works nicely.

I cannot comment on the code itself unfortunately but having looked through the files: I don't see anything that particularly stands out.

Approved! Thank you and well done.

TombEngine/Game/control/control.cpp
```diff
@@ -397,6 +397,7 @@ void KillMoveEffects()
 	ItemNewRoomNo = 0;
 }
 
+// NOTE: No one should use this ever again.
```
Jakub768, line 400: Maybe it is better to rename it to: "TODO: Remove this since it is useless and shouldn't be used" or something equivalent?

TombEngine/Game/items.cpp
```diff
@@ -182,13 +182,6 @@ bool ItemInfo::IsBridge() const
 	return Contains(BRIDGE_OBJECT_IDS, ObjectNumber);
 }
 
-bool ItemInfo::IsLaraCrouching() const
```
TokyoSU: Why removing it, it make the code more clean ?

Jakub768:
ok so I tested and I spotted 3 issues:

1) Flamethrower's flame does not affect the mutant, both of them try to to attack each other but neither gets damaged
2) mutant with OCB 0 (normal behaviour) tries to attack lara from far away and you have to move away considerably to make it move towards you
3) mutanbt trap version OCB 1 does not spit out the gas anymore (I think it is got to do with third argument - 8)

TokyoSU:
@Jakub768 Fixed the 3 :)

Sezzary (approved):

PR #1424, opened 2024-10-14
Replaced static limiting array for Flame Emitters to unlimited std::vector

goldlife2904:
## todo

(replace this sentence with links to issues this PR concerns, otherwise please write "n/a")


TombEngine/Game/effects/tomb4fx.cpp
```diff
@@ -371,31 +371,28 @@ void UpdateFireProgress()
...
+	Fires.push_back(newFire);
 }
 
 void ClearFires()
 {
-	for (int i = 0; i < MAX_FIRE_LIST; i++)
-		Fires[i].on = false;
+	for (auto& fire : Fires)
+	{
+		fire.on = false;
```
Lwmte, line 394: You are not clearing the vector, but instead just setting flag to false. It means that memory leak will occur.

TombEngine/Game/effects/tomb4fx.h
```diff
@@ -3,6 +3,7 @@
 #include "Game/Lara/lara_struct.h"
 #include "Math/Math.h"
 #include "Renderer/RendererEnums.h"
+#include <vector>
```
Lwmte, line 6: I am not sure if this include is needed?

Lwmte (changes requested):
Additionally, it may be good to reserve space for fires by using `Fires.reserve(DEFAULT_FIRE_COUNT)`, so that the vector is initially large enough, to avoid memory reallocation.

TombEngine/Game/effects/tomb4fx.h
Sylmir: Needed because the compiler needs to see a declaration of the type so that line 224 is valid C++

TombEngine/Game/effects/tomb4fx.cpp
goldlife2904: ah nice to know. Add the clear method.

TombEngine/Game/effects/tomb4fx.h
goldlife2904: it is not. Will change it.

TombEngine/Renderer/RendererDrawEffect.cpp
```diff
@@ -293,12 +292,12 @@ namespace TEN::Renderer
 
 	void Renderer::PrepareFires(RenderView& view) 
 	{
-		for (int k = 0; k < MAX_FIRE_LIST; k++) 
+		for (const auto& fire : Fires)
 		{
-			auto* fire = &Fires[k];
-			if (fire->on) 
+			
+			if (fire.on)
```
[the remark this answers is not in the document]
goldlife2904: ok, I will remove the flag.

TombEngine/Game/effects/tomb4fx.h
goldlife2904: as we cannot guarantee that it will always be included elsewhere I added it again.

TombEngine/Game/effects/tomb4fx.cpp
```diff
@@ -371,31 +371,26 @@ void UpdateFireProgress()
...
 {
-	FIRE_LIST* fptr = &Fires[0];
-	int i = 0;
-	while (fptr->on)
-	{
-		fptr++;
-		if (++i >= MAX_FIRE_LIST)
-			return;
-	}	
+	Fires.reserve(DEFAULT_FIRE_COUNT);
```
Lwmte, line 374: You can just remove `reserve` here :)

goldlife2904: done

TombEngine/Game/effects/tomb4fx.h
Sezzary: It's included by `framework.h`, so resolved by `tomb4fx.cpp`. You don't need it. ;)

TombEngine/Game/effects/tomb4fx.cpp
```diff
@@ -371,31 +371,25 @@ void UpdateFireProgress()
...
+	newFire.z = z;
+	newFire.roomNumber = roomNum;
+	newFire.size = size;
 
-	fptr->x = x;
-	fptr->y = y;
-	fptr->z = z;
-	fptr->roomNumber = roomNum;
-	fptr->size = size;
+	Fires.push_back(newFire);
```
Sezzary, line 387: The `Fires` vector shouldn't grow infinitely because effects need safeguards. Removing limits doesn't mean we literally remove them, rather we set *sensible* limits. :) If the legacy fire struct is still being used, you can write a special version of the `GetNewEffect()` function (found in `effects.h`) and call it here. When the vector has reached its maximum capacity and a new fire is being added, you could search for the smallest fire and overwrite it for example.

TombEngine/Game/effects/tomb4fx.h
goldlife2904: then I remove it again :-D

Lwmte (approved):

PR #1442, opened 2024-10-30
Load after death

Lwmte:
## Checklist

N/A

Alters the behaviour of game loop after player death:

- If any key except Esc key is pressed after death delay has passed, load game dialog is invoked.
- If Esc key is pressed after death delay has passed, game goes back to title, as always.

Load game dialog is only invoked if there are any valid savegames in any slot. Otherwise, game always goes back to title.

### DIFF
--- a/TombEngine/Game/control/control.cpp
+++ b/TombEngine/Game/control/control.cpp
@@ -631,7 +631,8 @@ GameStatus HandleMenuCalls(bool isTitle)
 
 	bool playerAlive = LaraItem->HitPoints > 0;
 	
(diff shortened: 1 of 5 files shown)


TombEngine/Game/control/control.cpp
```diff
@@ -90,6 +90,9 @@ using namespace TEN::Renderer;
 using namespace TEN::Entities::Creatures::TR3;
 using namespace TEN::Entities::Effects;
 
+constexpr auto DEATH_NO_INPUT_TIMEOUT = 10 * FPS;
+constexpr auto DEATH_INPUT_TIMEOUT	  = 3 * FPS;
```
goldlife2904, line 94: Is the space intended that way?

goldlife2904: ah, in VS it looks good, forget it :)

PR #1449, opened 2024-11-03
Adngel fix death camera

Adngel:
## todo

This PR addresses a camera bug occurring when Lara dies on a slope, causing a glitched, bouncing behavior. The issue originates from commit 53a7797.

That commit intended to fix camera behavior during special death animations. The camera code was refreshing the camera values on every frame, overwriting the specific values set for special death animations. To prevent this, that commit added a condition checking if Lara was alive before overwriting the Chase camera values. However, due to the lack of refreshing, this caused the default Chase camera to behave incorrectly when Lara died without triggering a special camera. (Like the camera bouncing when Lara dies on slopes).

This PR refines the conditions for camera refresh control. Now, it will ignore camera refreshing not only when Lara is dead but also when she is performing specific death animations.

### Key Changes:
- Added IsDeathCamera function to apply additional conditions during the Chase camera update, preventing interference with specific death animations.
- Configured unique camera angles and distances for spike and boulder deaths

### DIFF
--- a/TombEngine/Game/camera.cpp
+++ b/TombEngine/Game/camera.cpp
@@ -1289,7 +1289,10 @@ void CalculateCamera(const CollisionInfo& coll)
 	Camera.DisableInterpolation = (Camera.DisableInterpolation || Camera.lastType != Camera.type);
 	Camera.lastType = Camera.type;
 
-	if ((Camera.type != CameraType::Heavy || Camera.timer == -1) && !IsDeathCamera())
+	if (CalculateDeathCamera())
+		return;
+
+	if (Camera.type != CameraType::Heavy || Camera.timer == -1)
 	{
 		Camera.type = CameraType::Chase;
 		Camera.speed = 10;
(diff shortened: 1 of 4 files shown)


TombEngine/Game/camera.cpp
```diff
@@ -1304,6 +1304,45 @@ void CalculateCamera(const CollisionInfo& coll)
...
+		return false;
+	
+	// If Lara is in a special death animation (from extra_anims) triggered by enemies.
+	if (LaraItem->Animation.AnimObjectID == ID_LARA_EXTRA_ANIMS)
+		return true;
+
+	// Teeth spikes death animation
+	if (LaraItem->Animation.AnimNumber == LA_SPIKE_DEATH)
+	{
+		Camera.pos.RoomNumber = LaraItem->RoomNumber;
```
Lwmte, line 1320: I am not sure if it's nice idea to modify camera data inside a function called `IsDeathCamera`. It is a bad practice to change global variables in a function which is supposed just to check for certain conditions.

Adngel: Yeap, I took those commands away from the camera code in this next commit: [56665cd](https://github.com/MontyTRC89/TombEngine/pull/1449/commits/56665cd316a854d9c013ad4ac33f77308a7bd001)

Now they will be managed in Teeth_spike control code (After it activates Lara spikes death animation, it will set the camera values too)

And the same for boulder death

What leaves a simplified CalculateDeathCamera()  function only for the conditions check.

PR #1465, opened 2024-11-10
Expose wind

Joey79100:
## Checklist

N/A

This exposes to Lua the **wind vector** that was applied to the current frame by the engine.
This allows applying the same wind (or an altered version of it) to a custom object, or even Lara herself (for example if she's using a parachute).

I named the branch `expose_weather_parameters` in case other things related to weather come to mind (for example the wind's base strength, and direction, and maybe even overwriting them), but the PR is ready to use as-is otherwise.

_Feel free to amend the description in the doc, I'm not convinced myself with it!_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ TombEngine releases are located in this repository (alongside with Tomb Editor):
 	
 ### Lua API changes
 
+* Added Effects.GetWind() function.
+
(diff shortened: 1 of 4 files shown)


Jakub768:
Is it possible to have SetWind() too?

Joey79100:
> Is it possible to have SetWind() too?

No. The wind is question is the random vector computed by the engine to simulate the wind.
What could possibly be done is setting the base strength, and the direction (so they both need dedicated getters/setters). But I don't really understand which variable is what.

Lwmte (approved):

PR #1541, opened 2024-12-26
Multithreading

Sezzary:
## Checklist

n/a

Adds multithreading functionality to the codebase via the new `WorkerController` class.

Example usage:
```cpp
auto tasks = WorkerTaskGroup
{
    []() { MyTaskA(); },
    []() { MyTaskB(); },
    []() { MyTaskC(); },
    []() { MyTaskD(); }
};
g_Worker.AddTasks(tasks).wait();
```


TombEngine/Specific/Worker.h
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+	private:
+		// Helpers
+
+		void Worker();
+		void HandleTask(const WorkerTask& task, uint64_t groupId);
+	};
+
+	extern WorkerManager g_Worker;
```
Sylmir, line 78: Use the singleton pattern instead of an extern ? This will allow you to set the constructor to private and prevent other instanciations of this class

TombEngine/Specific/Worker.h
```diff
@@ -0,0 +1,79 @@
...
+				unsigned int threadCount = (itemCount > SERIAL_UNIT_COUNT_MAX) ? GetCoreCount() : 1;
+				unsigned int chunkSize = ((itemCount + threadCount) - 1) / threadCount;
+
+				// Handle task batches.
+				auto groupId = GetNewGroupId();
+				for (int i = 0; i < threadCount; i++)
+				{
+					int start = i * chunkSize;
+					int end = std::min(start + chunkSize, itemCount);
+					g_Worker.AddTask([task, start, end]() { task(start, end); }, groupId);
```
Sylmir, line 60: I'm reviewing this out-of-order, the use of g_Worker here is incredibly confusing (not to mention, I'm shocked the compiler allows you to reference an external you've not yet declared). If you make it clear WorkerManager is a singleton, you can remove the use of g_Worker (I first thought the function was static until I saw the call to GetNewGroupId, which is not static)

TombEngine/Specific/Worker.h
```diff
@@ -0,0 +1,79 @@
...
+		unsigned int GetCoreCount() const;
+
+		// Utilities
+
+		void AddTask(const WorkerTask& task, uint64_t groupId = (uint64_t)NO_VALUE);
+		void WaitForGroup(uint64_t groupId);
+		
+		// Template for batching parallel operations on vector.
+		template <typename T>
+		void ProcessInParallel(std::vector<T>& vec, const std::function<void(int, int)>& task, bool multiThreaded)
```
Sylmir, line 43: Consider returning a future to void instead of void so callers can wait on the future when they need to, instead of always having to wait for the completion of the group of tasks. This will further improve parallelism

You may also use a template template to have operations on any container, with your function taking a pair of iterators instead of a pair of indices (also, you should indicate as precondition that task should not change the underlying structure of the vector, otherwise the result of parallel execution is not the same as sequential execution)

TombEngine/Specific/Worker.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "Specific/Worker.h"
+
+namespace TEN::Utils
+{
+	WorkerManager g_Worker = WorkerManager();
+
+	WorkerManager::WorkerManager()
+	{
+		// Reserve threads.
+		unsigned int threadCount = std::thread::hardware_concurrency() * 2;
```
Sylmir, line 11: Replace the call to std::thread::hardware_concurrency with a call to GetCoreCount(), considering you already use that to alias std::thread::hardware_concurrency

TombEngine/Specific/Worker.h
Sylmir, line 60: I'm torn between creating a lambda and using std::bind_front to get a bind-expression, this sort of partial application to get a 0-arity function can be a bit hard to follow, even when you have a functional programming background IMO

TombEngine/Specific/Worker.cpp
```diff
@@ -0,0 +1,144 @@
...
+		_threads.reserve(threadCount);
+
+		// Create threads.
+		for (int i = 0; i < threadCount; i++)
+			_threads.push_back(std::thread(&WorkerManager::Worker, this));
+
+		_deinitialize = false;
```
Sylmir, line 18: Using operator= on an atomic variable induces std::memory_order_seq_cst which is probably not what you want here, especially when you consider that barring this access (which is not in parallel), all other accesses are under the protection of a mutex, which induces acquire-release ordering. You can probably let _deinitialize be a non atomic variable

TombEngine/Specific/Worker.cpp
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		// Notify all threads they should stop.
+		_taskCond.notify_all();
+
+		// Join all threads.
+		for (auto& thread : _threads)
+		{
+			if (thread.joinable())
+				thread.join();
```
Sylmir, line 37: Be careful, thread::join can throw, and destructors are not allowed to throw exceptions (it is UB to do so)

TombEngine/Specific/Worker.cpp
```diff
@@ -0,0 +1,144 @@
...
+				if (_deinitialize && _tasks.empty())
+					return;
+
+				// Get task.
+				task = _tasks.front();
+				_tasks.pop();
+			}
+
+			// Execute task.
+			if (task != nullptr)
```
Sylmir, line 116: std::function has an operator bool that does the same thing as this check, except it does not suggest task to be a pointer type. I think it's preferable to simply have if (task) here

TombEngine/Specific/Worker.cpp
```diff
@@ -0,0 +1,144 @@
...
+	void WorkerManager::Worker()
+	{
+		while (true)
+		{
+			auto task = WorkerTask();
+
+			// LOCK: Restrict task queue access.
+			{
+				auto taskLock = std::unique_lock(_taskMutex);
+				_taskCond.wait(taskLock, [this] { return (_deinitialize || !_tasks.empty()); });
```
Sylmir, line 104: Similar to what I suggested above, you should not have _definitialize be an atomic. Here, the conversion to bool through std::atomic<bool>::operator bool performs std::memory_order_seq_cst on the load operation, which is unnecessary strong considering you already locked a mutex when wait() woke up

TombEngine/Specific/Worker.h
```diff
@@ -0,0 +1,79 @@
...
+		{
+			constexpr auto SERIAL_UNIT_COUNT_MAX = 32;
```
Sylmir, line 45: Should be a configuration option IMO

TombEngine/Specific/Worker.h
```diff
@@ -0,0 +1,79 @@
...
+			
+			// Process in parallel.
+			unsigned int itemCount = (unsigned int)vec.size();
+			if (multiThreaded && itemCount > SERIAL_UNIT_COUNT_MAX)
+			{
+				unsigned int threadCount = (itemCount > SERIAL_UNIT_COUNT_MAX) ? GetCoreCount() : 1;
```
Sylmir, line 51: The condition always evaluates to true based on the condition of the if above on line 49

Sylmir, line 51: Also be careful, the standard allows GetCoreCount() to return 0, so the next line could possibly trigger a divide by zero

Sylmir (commented):
I only reviewed the threading part.

I don't see anything unsafe here, though there are two pairs of mutexes and condition variables (complexity) and at least one of them can go if you remove the group management with an atomic counter (my suggestion shown below). So I'm not 100% sure on the safety 

```cpp
template<typename T>
std::promise<void> ProcessInParallel(std::vector<T>& vector, std::function<int, int> const& fn)
{
    // ... Same as before
    std::atomic<int> counter;
    counter.store(threadCount, std::memory_order_release);
    
    std::promise<void> promise;
    for (int i = 0; i < threadCount; ++i)
    {
        int start = i * chunkSize;
        int end = std::min(start + chunkSize, itemCount);
        AddTask([fn, start, end] () -> void { fn(start, end); }, counter, promise);
    }
    
    return promise;
}

void WorkerManager::HandleTask(const WorkerTask& task, std::atomic<int>& counter, std::promise<void>& promise)
{
    task();
    
    if (counter.fetch_sub(1, std::memory_order_acq_rel) == 1)
        promise.set_value();
}
```

Sezzary:
@Lwmte I've heavily modified your template method following Sylmir's review. Unless you had another idea in mind for it and you think it needs further revision, the PR is ready.

TombEngine/Renderer/RendererSprites.cpp
```diff
@@ -414,11 +419,11 @@ namespace TEN::Renderer
 			_shaders.Bind(Shader::InstancedSprites);
 
 			// Set up vertex buffer and parameters.
-			UINT stride = sizeof(Vertex);
-			UINT offset = 0;
+			unsigned int stride = sizeof(Vertex);
```
Lwmte, line 422: I am not sure if it's a good idea to do that. This data type is derived from the arguments of `GetAddressOf`.

TombEngine/Renderer/RendererSprites.cpp
```diff
@@ -261,16 +263,16 @@ namespace TEN::Renderer
 			return;
 
 		// Draw instanced sprites.
-		bool wasGPUSet = false;
-		for (auto& spriteBucket : _spriteBuckets)
+		bool wasGpuSet = false;
```
Lwmte, line 266: Not sure if it should be Gpu, because GPU is not a name, it is an abbreviation.

Sezzary: Abbreviations longer than two characters are still treated as words by convention. In fact I think even two-letter abbreviations should be treated like regular words too because of some oddities you can get, but that's another matter. :)

PR #1583, opened 2025-02-25
Horizon Effects

TrainWrack:
## Checklist

(replace this sentence with links to issues this PR concerns, otherwise please write "n/a")

Adds 3 functions

SetHorizonRotation(Rotation())
GetHorizonRotation(); Returns Rotation()
SetHorizon(ObjectSlot)

Sample Script:

```
local speed = 0.0
local test = false
local increment = 0

LevelFuncs.OnLoop = function()
 
local rotation = GetHorizonRotation()

  local myTextString = tostring("Rotation: " .. rotation.y)
  
  local myText = DisplayString(myTextString, 100, 400, Color.new(64,250,60))
  ShowString(myText,1/30)
  
  if test then
      speed = (speed + increment)%360
      SetHorizonRotation(Rotation(rotation.x,speed,rotation.z))
  end
  
  if test2 then
      speed = (speed + increment)%360
      SetHorizonRotation(Rotation(speed,rotation.y,speed))
  end

end

LevelFuncs.Rotate1 = function() 
  test = true
  increment = 1
end


LevelFuncs.Rotate2 = function() 
  test = false
  test2 = false
  SetHorizonRotation(Rotation(0,0,0))
end


LevelFuncs.Rotate3 = function() 
    test2 = true
    increment = .1
end
LevelFuncs.Swap1 = function() 
  SetHorizon(TEN.Objects.ObjID.MESHSWAP1)
end


LevelFuncs.Swap2 = function() 
    SetHorizon(TEN.Objects.ObjID.MESHSWAP2)
end


LevelFuncs.Swap3 = function() 
    SetHorizon(TEN.Objects.ObjID.HORIZON)
end
```


TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
```diff
@@ -42,6 +43,11 @@ using namespace TEN::Scripting::Types;
 
 namespace TEN::Scripting::Effects
 {
+	// Initialize the global variables for Horizon Rotation and Swap
+	Vector3 _horizonRotation = Vector3::Zero;
```
Lwmte, line 47: I am not sure it's a good idea to make these variables global. I suggest to move it to `Level` class, which already contains parameters for fog, sky layers and draw distance.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
```diff
@@ -371,9 +401,13 @@ namespace TEN::Scripting::Effects
 		tableEffects.set_function(ScriptReserved_EmitFire, &EmitFire);
 		tableEffects.set_function(ScriptReserved_MakeEarthquake, &Earthquake);
 		tableEffects.set_function(ScriptReserved_GetWind, &GetWind);
+		tableEffects.set_function("SetHorizonRotation", &RotateHorizon);
```
Lwmte, line 404: Function names shouldn't be directly specified in register call. Move them to `ScriptReserved_` constants.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
TrainWrack: Addressed

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
TrainWrack: Addressed

TombEngine/Game/effects/weather.cpp
```diff
@@ -21,6 +21,29 @@ using namespace TEN::Math;
 
 namespace TEN::Effects::Environment 
 {
+	GAME_OBJECT_ID HorizonObject::_horizonID = ID_HORIZON;
```
Lwmte, line 24: I think that inintialization of default values should be within a class, not outside it.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3021,7 +3024,11 @@ namespace TEN::Renderer
 		}
 
 		// Draw horizon.
-		if (_moveableObjects[ID_HORIZON].has_value())
+		TEN::Effects::Environment::HorizonObject horizonC;
```
Lwmte, line 3027: Why do you declare new `horizonC` class value here? It should be in a script environment, unless I am missing something?

TombEngine/Renderer/RendererFrame.cpp
```diff
@@ -901,6 +903,10 @@ namespace TEN::Renderer
 				item.PrevAnimTransforms[j] = item.AnimTransforms[j];
 		}
 
+		//SAVE HORIZON INTERPOLATION DATA
+		TEN::Effects::Environment::HorizonObject horizon;
```
Lwmte, line 907: Same as above. You shouldn't use static members within a class and then define a class value in two separate places, in my opinion.

TombEngine/Game/effects/weather.cpp
TrainWrack: This is what GPT Chan said.

Why Can't Static Variables Be Initialized in the Header?
Storage Allocation Issue

A static member variable belongs to the class itself (not any specific object), so it requires a single definition in a .cpp file to allocate memory for it.
If you initialize it inside the class definition, the compiler doesn't know where to store it.
Multiple Definition Error in Multiple Translation Units

If you define the variable inside the header file, every .cpp file that includes the header will get its own separate copy

I need the variables to be static as they are being used across the files.

I could be wrong as I am not a programmer by any means.

TrainWrack: Addressed

TombEngine/Renderer/RendererDraw.cpp
TrainWrack: Addressed

TombEngine/Renderer/RendererFrame.cpp
TrainWrack: Addressed

TrainWrack:
Tested the branch. Savegame and rotation saved perfectly. Reloaded the level and it restored. 

Restarted the level and it was defaulty horizon and rotation. Played another level and it loaded the default horizon as well.

PR #1615, opened 2025-03-15
Advanced Particles II

TrainWrack:
## Checklist

N/A

Adds another field called drawRot which accepts a rotation value. The particle is drawn based on the rotation specified. This allows to create flat particles that do not pivot to face camera.


TombEngine/Renderer/RendererSprites.cpp
```diff
@@ -103,6 +103,38 @@ namespace TEN::Renderer
 		view.SpritesToDraw.push_back(spr);
 	}
 
+	void Renderer::AddSpriteBillboardRotated(RendererSprite* sprite, const Vector3& pos, const Vector4& color, float orient2D,
```
Lwmte, line 106: Can you explain why do you need one extra `RotatedBillboard` sprite type and render function? To me, this looks like ordinary constrained axis sprite. If you need free axis instead of constraining to a particular X/Y/Z axis, then constrained sprite type should be modified instead of writing additional sprite type.

TrainWrack: Constrained axis type billboards are only constrained on one axis. These ones are constrained on all 3. Therefore we can truly have flat particles that are rotated at the specified rotation.

Lwmte: I understand. What I am suggesting is to modify constrained sprite to support all 3 axes, and not creating another type.

TrainWrack: I thought about this and its a bit difficult to merge the two. 

LookAtBillboard is view-aligned (camera-facing). The sprite rotates to face the camera all the time while being locked on an axis.

RotatedBillboard uses raw rotation angles (orientation).  These are drawn based on the rotation vector. They never face the camera. 

Both of them have different world matrix calculations also.

Not sure how to begin merging the two. Open to suggestions.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
```diff
@@ -398,6 +399,14 @@ namespace TEN::Scripting::Effects
 			if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WIND, part.roomNumber))
 				part.flags |= SP_WIND;
 		}
+
+		Rotation convertedConstraint = table.get_or("drawRot", Rotation(0, 0, 0));
+
+		if (!(convertedConstraint == Rotation(0, 0, 0)))
```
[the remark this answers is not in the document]
TrainWrack: Not relevant

TombEngine/Renderer/RendererSprites.cpp
Lwmte: I checked again. Type `CustomBillboard` is used for many types of oriented sprites - lasers, meteors, electricity sparks, ricochets etc. How is that feature differs from what you are suggesting?

TrainWrack:
@Lwmte Can we finalize the approach for these particles? I would need fixed rotation particles also for lua particles so if this gets merged I can use the same approach.

TombEngine/Renderer/RendererSprites.cpp
TrainWrack: I have removed the new sprite type and utilized addquad

Stranger1992 (approved):
Approved from other user feedback. All comments have been addressed and the only query I had (regarding the new `Haze` blend type was addressed on discord and is included.

PR #1625, opened 2025-04-02
Video playback

Lwmte:
## Checklist

N/A

Adds video playback support via libVLC. Can be called from the console or by script command `PlayVideo("fileName")` where `fileName` is the name of a file in `FMV` subdirectory (similar to classic games).

Playback requires additional istallation of VLC plugins from VLC SDK:

- Download VLC SDK from here (x64): https://artifacts.videolan.org/vlc/nightly-win64-llvm/
- Unzip it and copy `/plugins` directory to TEN's `Bin/x64` directory
- Place some test video file into `/FMV` directory, e.g. `test.mp4`
- Type in the console `PlayVideo("test")`

As in classic games, video will interrupt current game, including sounds. Video can be stopped by pressing Esc or Look.


TombEngine/Specific/Video/Video.cpp
```diff
@@ -0,0 +1,377 @@
...
+			player->_d3dContext->Unmap(player->_videoTexture, 0);
+			player->_needRender = true;
+		}
+		else
+		{
+			TENLog("Failed to map video texture", LogLevel::Error);
+		}
+	}
+
+	void VideoHandler::OnDisplayFrame(void* data, void* picture)
```
Jakub768, line 344: I guess this is a function that needs to be implemented from Video handler class?

Jakub768 (approved):
Code looks good, needs testing.

TombEngine/Specific/Video/Video.cpp
Lwmte: Thank you for noticing, I have removed it, as placeholder callbacks are not needed.

PR #1638, opened 2025-04-22
issue 1584, attempt to fix clipping into angled floors when entering the LS_LADDER_TO_CROUCH state from climbing (discord: bard, Jordan B)

jorbar-Dev:
## Checklist

https://github.com/TombEngine/TombEngine/issues/1584 

two changes:

climb idle state transfers to upward climb state instead of LS_LADDER_TO_CROUCH when attempting to climb onto a flat / angled floor, so that upward climb can handle this collision

upward climb state now requires both probes to detect if there is no wall present
- LS_LADDER_TO_CROUCH should only occur if near edge of wall
- player can keep moving upwards otherwise


CHANGELOG.md
```diff
@@ -3,8 +3,14 @@
 The dates are in European standard format where date is presented as **YYYY-MM-DD**.
 TombEngine releases are located in this repository (alongside with Tomb Editor): https://github.com/TombEngine/TombEditorReleases
 
+
 ## [Version 1.8.2]
 
+### changes
```
Lwmte, line 9: Entries in the changelog should be either in "Bug fixes" or "New features" section - I think that in this case, all three entries can go to "Bug fixes" section. Also, in my opinion, it is not necessary to provide such detailed changelog entries as first two - you can combine it to an entry like "Fixed issues with ladder climbing near slanted ledges".

TombEngine/Game/Lara/lara_climb.cpp
```diff
@@ -344,8 +347,14 @@ void lara_col_climb_idle(ItemInfo* item, CollisionInfo* coll)
 			}
 			else
 			{
-				item->Animation.TargetState = LS_LADDER_TO_CROUCH;
-				item->Animation.RequiredState = LS_CROUCH_IDLE;
+				// Hack: disable this state transition to avoid clipping bug on issue 1584.
```
Lwmte, line 350: I am not sure how this change fixes ladder to crouch issue, if it completely disables such state change? Does `lara_col_climb_up` automatically sort it out?

TombEngine/Game/Lara/lara_climb.cpp
```diff
@@ -155,7 +155,10 @@ void lara_col_climb_up(ItemInfo* item, CollisionInfo* coll)
 		 
 		if (IsHeld(In::Forward) && resultRight && resultLeft)
 		{
-			if (resultRight < 0 || resultLeft < 0)
+			// Hack: to avoid clipping into angled wall, make sure both probes are not detecting the wall
```
Lwmte, line 158: Maybe I am wrong, but isn't `||` operation more restrictive than `&&` in this case? If you change to `&&`, it means that both conditions must succeed, not just one.

Issue #1334, opened 2024-03-03
Tomb Engine Requested Features


Stranger1992:
It was recently asked on the TombEngine discord server what features users request. Rather than making a separate issue for each: I have collated them here.

# Gameplay Mechanics / Level Creation.
- [ ] A more flexible save / load from PSX.
- [ ] All TR1-TR5 objects and features.
- [ ] Allow different weapon types. (Different firing types, different damages and such ) .
- [ ] Artificial-Inteligence overhaul..
- [ ] Assigning faction/allegiances to entities ( change how they act towards lara, friendly or not etc ) .
- [ ] Batch Import of Imported Geometry.
- [ ] Better handling of audio / sfx .
- [ ] Diary system from TRNG .
- [ ] Export models with animations from .wad2 with WadTool for Blender / 3DS Max .
- [ ] Force animation from different slot on Lara.
- [ ] Free axis rotation on statics.
- [ ] Glide timer for TEN cameras.
- [ ] Hub System.
- [ ] In-Depth Combat / Stealth system.
- [ ] More enemy parameters..
- [ ] TR5's Keypad / Panel Puzzles in TEN .
# Lua API.
- [ ] Better handling of audio / sfx.
- [ ] Editable Camera via LUA.
- [ ] Editable Flares / Personal Light (colour, light flicker parameters, duration).
- [ ] Interact with objects in the inventory.
# Rendering and Graphics.
- [ ] 60 FPS..
- [ ] Allow fire/smoke to have a different texture .
- [ ] Configurable mip-mapping distance .
- [ ] Denser weather effects - more rain and snow particles..
- [ ] Enemy fade-outs after killing. .
- [ ] FMV'S / MP4 implementation .
- [ ] Fade In / Fade out effect / Displaying one letter at a time for Strings.
- [ ] Fix Young Lara hair (IE: twin pigtails).
- [ ] Larger textures per sector in Tomb Editor (ie: 512px / 1024px / 2048px).
- [ ] Parallax mapping and UV off setting.
- [ ] Remaining legacy effects (lens flares, mirrors, pseudo-cubemaps like TR1-2 reflections).
- [ ] Ring Inventory Menu (in the style from Tomb Raider I - III ).
- [ ] Separate texture for the water splash texture + splash ring.
- [ ] White gloss overlay on status bars optional/editable.
- [ ] ambient cubemaps and lightmapping.
- [ ] fix any ongoing bugs (e.g. gun smoke bug).
[...]

PR #1643, opened 2025-04-25
Clustered weather

Lwmte:
## Checklist

#1334 

Implements much more dense weather effect by implementing particle clustering. Additionally solves some particle collision issues and adds random cluster rotation of snow particles around root particle, making snow effect more natural.


Stranger1992 (approved):
## Testing

### Rain

Rain is now much more...dense if that is the right word. Setting the strength to 1.0 produces an almost bad-tv effect on screen that looks kind of unconvincing. But that also may be because it is at full strength. However this is a comment and not necessarily part of this PR's purpose.

I have created a WAD2 with rain sprites that mitigates this to some degree.
[rain.zip](https://github.com/user-attachments/files/19929001/rain.zip)

I also noticed that, when going from a camera view to the chase camera: the rain will stop for a second before restarting.  

- [x] Works when flipping the room
- [x] Rain strength is saved and retained on save/load sequence
- [x] Rain does not generate ripple sprites on quicksand surface. NB: This may be the case originally but thought I'd mention it.

### Snow

The snow is looking great! It fades and rotates naturally. 

I repeated the same tests with the rain: 
- [x] Works when flipping the room
- [x] Snow strength is saved and retained on save/load sequence
- [x] Snow settles on a quicksand surface

I cannot directly comment on the code side: but these are my findings. Take these suggestions on a basis that they are optional. I will approve this PR but requires a code review.

PR #1651, opened 2025-05-03
Better input conflict management + expose raw keyboard and mouse input

Sezzary:
## Checklist

- [ ] I have added a changelog entry to CHANGELOG.md file on the branch/fork (if it is an internal change then it is not needed)

N/A

Came up with a relatively simple solution to our problems. My older branch, `sezz/input-action-groups`, was introducing too much complex polymorphism and C++ nonsense. In this PR, I settled for something more C-ish and maintainable.

- BUG FIX: Handle binding conflicts correctly. Different binding groups (general, vehicle, menu, etc.) no longer interfere when two input actions from different action groups are mapped to the same key.
- Allow querying raw keyboard and mouse input.
- Expose raw keyboard and mouse action keys in Lua.'
- Revise Lua input functions.
- Add new Lua function `IsKeyReleased`, which returns `true` if the user has just let go of a key. Has optional `maxDelaySec` arg (check doc comments for details).
- Add new Lua function `IsKeyPulsed`, which returns `true` at specified time intervals while the user holds down a key. Has optional `initialDelaySec` arg (check doc comments for details).
- Add new Lua function `GetAnalogKeyValue`, which returns a value in the range `[0, 1]` for keys mapped to analog inputs or simply 0/1 for discrete digital inputs.
- Update `IsKeyHeld` Lua function with an optional `delaySec` arg (check doc comments for details).

TO TEST:
- Check all new additions to key query functions above.
- Check if queries with ALL keyboard and mouse actions work correctly. For an extensive list, look at `ActionIDs.h` in the code section of this PR. For simplicity, only common keys present on most keyboards and mice are supported. No numpad keys or extra mouse buttons.
- Check if `SCROLL_UP` and `SCROLL_DOWN` actually correspond to scrolling up and down with the mouse. They may be flipped.
- Check if mapping two input actions from different action groups (general, vehicle, quick, and menu) to the same key results in interference.

TO DO:
- Nodes with action ID drop-down menus will be overloaded with a GIANT list of actions because they all belong to a single enum. Hopefully, we can come up with a clean solution. If viable, I was thinking two drop-down menus: the first to select the action ID group (`General`, `Vehicle`, `Quick`, `Menu`, `Keyboard`, `Mouse`) and the second to select the actual action ID from that group.


TombEngine/Scripting/Internal/TEN/Input/ActionIDs.h
```diff
@@ -1,112 +1,240 @@
 #pragma once
-#include "Specific/Input/InputAction.h"
 
-#include <string>
-#include <unordered_map>
+#include "Specific/Input/InputAction.h"
 
 using namespace TEN::Input;
 
-/***
```
Lwmte, line 9: Just a suggestion: maybe in the future we may write a PowerShell script to autogenerate this enumeration in a similar way to `GAME_OBJECT_ID` enum script which creates corresponding lua enum.

TombEngine/Scripting/Internal/TEN/Input/InputHandler.cpp
```diff
@@ -37,61 +37,82 @@ namespace TEN::Scripting::Input
...
 			return false;
 
-		if (IsHeld((InputActionID)actionID))
-			return true;
+		return IsClicked((ActionID)actionID);
+	}
+
+	/// Check if an action key is being held.
+	// @function IsKeyHeld
+	// @tparam float[opt=0] delaySec Delay time in seconds before a hold can be registered.
```
Lwmte, line 53: Please change order of `tparam` here to correspond to actual function signature.

TombEngine/Scripting/Internal/TEN/Input/InputHandler.cpp
```diff
@@ -37,61 +37,82 @@ namespace TEN::Scripting::Input
...
 			return false;
 
-		if (IsClicked((InputActionID)actionID))
-			return true;
+		return IsPulsed((ActionID)actionID, delaySec, ValueOr<float>(initialDelaySec, 0.0f));
+	}
+
+	/// Check if an action key is being released.
+	// @function IsKeyReleased
+	// @tparam float[opt=infinity] maxDelaySec Max delay time in seconds between hit and release within which a release can be registered.
```
Lwmte, line 79: As above.

TombEngine/Specific/configuration.cpp
```diff
@@ -269,22 +277,26 @@ bool SaveConfiguration()
 	}
 
 	if (g_Configuration.Bindings.empty())
-		g_Configuration.Bindings = BindingManager::DEFAULT_KEYBOARD_MOUSE_BINDING_PROFILE;
+		g_Configuration.Bindings = DEFAULT_KEYBOARD_MOUSE_BINDING_PROFILE;
 
 	// Set Input binding keys.
-	for (int i = 0; i < (int)InputActionID::Count; i++)
+	for (auto actionGroupID : SAVEABLE_ACTION_GROUP_IDS)
```
Lwmte, line 283: If I understand correctly, now every action group is saved separately?

Sezzary: This is supposed to ignore hardcoded keyboard/mouse bindings and do the same as before.

Just note I made an error somewhere and saving the config doesn't work ATM.

Sezzary: Binding config loading's been fixed now.

PR #1693, opened 2025-07-06
Rework enum documentation - use standardised LDoc implementation

Nickelony:
## Checklist
- [x] ~~I have added a changelog entry to CHANGELOG.md file on the branch/fork (if it is an internal change then it is not needed)~~

n/a

Reworked enum documentation - using standardised LDoc implementation.
Also added documentation for `LaraType` and `WeatherType`.

All 4 PRs combined can be tested on this branch: https://github.com/TombEngine/TombEngine/tree/Nickelony/LDoc/Testing


TombEngine/Scripting/Internal/TEN/Input/ActionIDs.h
```diff
@@ -7,246 +7,451 @@ using namespace TEN::Input;
...
+		// @mem MOUSE_SCROLL_UP
 		{ "MOUSE_SCROLL_UP", In::MouseScrollUp },
-		{ "MOUSE_SCROLL_DOWN", In::MouseScrollDown }
+
+		/// Mouse scroll down.
+		// @mem MOUSE_SCROLL_DOWN
+		{ "MOUSE_SCROLL_DOWN", In::MouseScrollDown },
+
+		// COMPATIBILITY
+		{ "1", In::Num1 },
```
Lwmte: I am not sure we need this compatibility section, given that these never worked in the first place?

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
```diff
@@ -126,7 +126,7 @@ namespace TEN::Scripting::Effects
 	// @tparam[opt=2] float life Lifespan in seconds.
 	// @tparam[opt=false] bool damage Harm the player on collision.
 	// @tparam[opt=false] bool poison Poison the player on collision.
-	// @tparam[opt=Objects.ObjID.DEFAULT_SPRITES] Objects.ObjID.SpriteConstants spriteSeqID Sprite sequence slot ID.
```
Lwmte, line 129: Why `SpriteConstants` namespace was removed? The point of this namespace was to define sprite object IDs not to confuse builders with using non-sprite object IDs for this family of functions.

Documentation/generate_objectlist.ps1
```diff
@@ -73,57 +73,26 @@ $header = @"
...
-@table SpriteConstants
-*/
-"@
-
 # Map definition.
 $mapHeader = "static const std::unordered_map<std::string, GAME_OBJECT_ID> GAME_OBJECT_IDS {"
-$mapBody = ($enumValues | ForEach-Object { "`t" + '{ "' + "$_" + '", ID_' + "$_" + ' }' }) -join ",`r`n"
+$mapBody = ($enumValues | ForEach-Object {
+    $enumName = $_
+    $description = "Generic constant"
```
Lwmte: Can we change description somehow, so that it doesn't show just "Generic constant" but at least something more descriptive? Like "Object ID"?

TombEngine/Scripting/Internal/TEN/View/DisplaySprite/ScriptDisplaySprite.cpp
```diff
@@ -52,7 +52,7 @@ namespace TEN::Scripting::DisplaySprite
 
 	/// Create a DisplaySprite object.
 	// @function DisplaySprite
-	// @tparam Objects.ObjID.SpriteConstants objectID ID of the sprite sequence object.
```
Lwmte, line 55: As stated above.

TombEngine/Scripting/Internal/TEN/Input/ActionIDs.h
Nickelony: It was still possible to access those keys, but you had to use the dictionary syntax.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
Nickelony: The namespace didn't actually exist, it's never defined anywhere else.

Documentation/generate_objectlist.ps1
Nickelony: This entire file is automatically generated using a power shell script. I'm not sure if adding custom descriptions is possible with the existing approach. I added the distinction between Generic and Sprite constant to somehow preserve the namespace intentions.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
Lwmte: It did exist. You just removed a part which generated it from PowerShell script. Same goes for `PickupConstants`.

Documentation/generate_objectlist.ps1
Lwmte: Not custom descriptions, I mean just changing "Generic constant" to something more readable for builders, like "Generic object ID", "Sprite object ID", "Pickup object ID".

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
Nickelony: We need to split these into their own separate enums then, as LDoc itself doesn't recognise these as valid types / modules. Not sure what else to do in this case. They are in-code all part of the same `ObjID` enum, whereas the `SpriteConstants` notation is not recognized by the C++ code. I assume this only works since the calls are being shortened? So the part before the `.` is completely removed? Otherwise, this is completely invalid syntax.

Documentation/generate_objectlist.ps1
Nickelony: Yea, definitely can do! I will push a fix later.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
Lwmte: Yeah, probably something overlooked by @davidmarr.

davidmarr: The three tables were used to provide a visual grouping, solely to assist the builders.

Lwmte: @Nickelony can we introduce something like that? group object IDs to particular subgroups and reference them with autocomplete?

TombEngine/Scripting/Internal/TEN/Input/ActionIDs.h
Lwmte: I doubt anyone ever used that, or even had any idea it could have been used this way, so I am sure it's safe to remove it.

TombEngine/Scripting/Internal/TEN/Effects/EffectsFunctions.cpp
Nickelony: Not without splitting these into 3 separate enums I'm afraid. :(

PR #1769, opened 2025-11-06
Switch to faster Xorshift32 random number generator instead of mt19937

Lwmte:
## Checklist

- [ ] I have added a changelog entry to CHANGELOG.md file on the branch/fork (if it is an internal change then it is not needed) 

N/A

Switches to Xorshift32 random number generation method, which is 15-30x faster than currently used method:

Clipboard-1

### DIFF
--- a/TombEngine/Game/savegame.cpp
+++ b/TombEngine/Game/savegame.cpp
@@ -1162,6 +1162,8 @@ const std::vector<byte> SaveGame::Build()
 	auto* level = (Level*)g_GameFlow->GetLevel(CurrentLevel);
 	Save::LevelDataBuilder levelData { fbb };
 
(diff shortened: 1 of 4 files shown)


Lwmte:
> Something to note is that a new game start will be entirely predictable every time. It may be a good idea to start with a truly random seed at the beginning of a new game, though how this is implemented depends on the needs of demo replays.

For now it's not a big concern, because in most cases, new game will still diverge to different behaviour. This is caused by us using `Random::GenerateInt()` inconsistently in both rendering and game logic code. Core used two seeds (via `GetRandomControl` and `GetRandomDraw`) to solve this problem. For the same reason, rolling demos are impossible to implement unless we convert `Random::GenerateInt` usage back to `GetRandomControl` in control logic code and implement separate seed for the game logic.

PR #1785, opened 2025-11-20
Remove forced camera for turn switch

JoeyQuint:
## Checklist

_N/A_

In TR4, using the turn switch activates a fixed camera roughly above the switch, which tries to follow Lara. This camera is completely hardcoded.
The goal was probably to see the orientation of the obelisk below (that the switch controls in TR4).
Except:
- The builder cannot override it with their own camera placed in-editor. In TR4, it would just flicker, and in TEN it's simply ignored.
- In TR4, it tried to follow her but there was a snap at the end, and in TEN it just remains stuck at her initial position, until it loops again.

TR4 behavior (just the hardcoded camera):

https://github.com/user-attachments/assets/b4bba0dd-6243-4a9f-865b-89af9e667246

TEN behavior, with a camera placed in editor triggered around the switch:

https://github.com/user-attachments/assets/ba5fa0aa-d057-4400-824e-33d47df37add

This PR removes this hardcoded camera and makes it follow Lara instead as she moves. This looks more natural, and most importantly opens the ability to place a fixed camera in-editor instead.
If the builder really wants the camera only while Lara is pushing the switch, they can replicate that using a volume, which would trigger the camera only when Lara's current state is 95 (used by all the turn switch animations).

New result with no camera:

https://github.com/user-attachments/assets/768ae05b-bd2b-418c-b957-975d0fdf8ad1

New result with a camera placed in-editor (Lara being the target):

https://github.com/user-attachments/assets/bd66a408-d742-42f4-be10-50c5e65e8f3f

New result with a camera placed in-editor (with a fixed camera target):

https://github.com/user-attachments/assets/130a4582-d0f9-43a4-9a45-abeabaee10ec


TombEngine/Game/Lara/lara_basic.cpp
```diff
@@ -83,6 +83,18 @@ void lara_as_controlled_no_look(ItemInfo* item, CollisionInfo* coll)
 	coll->Setup.EnableSpasm = false;
 }
 
+void lara_as_controlled_no_look_follow(ItemInfo* item, CollisionInfo* coll)
```
Lwmte, line 87: Do we need a new state function for this? Can we reuse existing ones?

JoeyQuint: It was originally using `lara_as_controlled_no_look`, but:
- I needed `Camera.flags = CF_FOLLOW_CENTER` to get the camera moving with Lara
- I also needed `Camera.laraNode = LM_HEAD` to make it rotate, in this case following her head (which was the most stable mesh - that's because the default anim is weird and has her twisting left and right)
- And `Camera.targetElevation = -ANGLE(25.0f)` moves the camera up, which avoids the camera fighting with the switch's collision box - and it also looks better overall

An alternative could be using `lara_as_pushable_push`, which is the closest candidate, but:
- It has the camera tilted to the left side (for the turnswitch it could work, but only when pushing clockwise)
- It won't rotate as Lara rotates (it will just move forward, and then snap at the end of 1 turn)
- The name implies this is reserved for pushable objects.
  We're technically not so far from it (Lara is pushing an object), but it could lead to unwanted side-effects if changes in the future were done specifically for pushable objects to this function - and pushables are special objects with very specific handling that probably "deserve" their dedicated functions to avoid this happening?

So I thought it was reasonable enough to make a new one - generic enough because there isn't much to it, and it could possibly fit other "simple" objects.

PR #1801, opened 2025-11-27
Include {fmt} library

Sezzary:
## Checklist

- [x] Pull request meets the Coding Conventions standards:https://github.com/MontyTRC89/TombEngine/blob/master/CONTRIBUTING.md#coding-conventions

n/a

A quick PR that includes the {fmt} library in the framework. This was available to use the whole time because {fmt} is packaged with spdlog.

Before:
```C++
TENLog("Logging error message with C string " + std::string(myCString) + ", an integer " + std::to_string(myInteger) + ", and another integer for good measure: " + std::to_string(myOtherInt) + ".", LogLevel::Error);
```

After:
```C++
TENLog(Fmt("Logging error message with C string {}, an integer {}, and another integer for good measure: {}.", myCString, myInt, myOtherInt), LogLevel::Error);
```

`Fmt` is a shorter alias for the function `fmt::format`. As this call will be used extensively, it's nice to have for brevity, and I've found it comfortable in my own projects. However, we can still use the full `fmt::format` name if preferred.

### DIFF
--- a/TombEngine/framework.h
+++ b/TombEngine/framework.h
@@ -56,8 +56,6 @@ using namespace DirectX::SimpleMath;
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/stdout_color_sinks.h>
 
-#define Fmt fmt::format
-
 // VLC
 #include <vlc/vlc.h>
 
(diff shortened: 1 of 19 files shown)


TombEngine/framework.h
```diff
@@ -48,6 +50,14 @@ using namespace DirectX::SimpleMath;
 // sol
 #include <sol.hpp>
 
+// spdlog
+#include <spdlog.h>
+#include <spdlog/fmt/fmt.h>
+#include <spdlog/sinks/basic_file_sink.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+
+#define Fmt fmt::format
```
Lwmte: Maybe it's better to use `Format` word, and also put this under `TEN::Utils` namespace in `trutils.cpp/h` file? This shouldn't really sit in the framework.h in my opinion.

Issue #1810, opened 2025-11-29
SEARCH_OBJECT4 sometimes doesn't "find the pickup"


Joey79100:
### Tomb Engine Version

Development build

### Development Version

- [ ] I am using an unofficial development version.
- [x] I am using an official release.
- [ ] I am using an official pre-release.

### Describe the Bug

Sumi Beans reports SEARCH_OBJECT4 (the chest from TR5 VCI) doesn't work for her.

>   searchable chest, it works but doesnt trigger the item, if you put an item in front of the searchable, the item can only be triggered with that object, but the chest isnt doing this, the other 3 work fine.
> <img width="888" height="652" alt="Image" src="https://github.com/user-attachments/assets/71d1703f-67ad-4455-b201-80dfd792c976" />
> no, the other 3 searchable objects have their item at different heights and work fine

What happens: Lara plays the pickup animation, the chest also plays it, but the item is never picked.
When switching to wireframe mode, we can see the item actually is visible the entire time, and remains so. It seems the SEARCH_OBJECT4 doesn't detect it.

After testing myself, I can confirm that for some reason, it works on a project of mine, but not in another.
The setup is the exact same, the objects are not.

### To Reproduce

Place a SEARCH_OBJECT4 and a pickup item on the same position.
- If all is fine, Lara should pick the item from the chest.
- But if all is not fine, she will just play the animation without the item being picked up.

### Expected Behaviour

If a pickup item is placed at the same location as a SEARCH_OBJECT4:
- The item should turn invisible
- Lara should manage to pick it up when playing the chest animation

### Minimal Reproduction Project

Example project where it doesn't work, by Sumi Beans on Discord:
- Main archive: https://drive.google.com/file/d/1-O4Zhpmmjg_fitlKWjssMAGX-OHtaPox/view?usp=sharing
- Wad: https://drive.google.com/file/d/1BK7Kyg0gQkbiUBWO8ngv1bZQWsUhZaF4/view?usp=sharing

Sumi-Beans:
this doesnt seem to be fixed to me, unless im still using the old tomb engine, i did upgrade my tomb engine, with what was given to me.
[...]

PR #1811, opened 2025-11-29
Fix and improve SEARCH_OBJECT4

JoeyQuint:
## Checklist

#1810 

Fixes SEARCH_OBJECT4 not always collecting the pickup item sitting under it.  
See #1810 for details on what was causing the issue.

Also makes it work like other SEARCH_ITEMs under the hood, rather than using its own specific code _(done as a separate commit for clarity)_:
- Allows for multiple items rather than just one
- Makes the item invisible (necessary in case they are too big for the chest).

### DIFF
--- a/TombEngine/Specific/level.cpp
+++ b/TombEngine/Specific/level.cpp
@@ -1901,7 +1901,7 @@ void GetCarriedItems()
 					abs(item2.Pose.Position.z - item.Pose.Position.z) < CLICK(2) &&
 					abs(item2.Pose.Position.y - item.Pose.Position.y) < CLICK(1))
 				{
-					bool isSearchObjectWithExplosion = (item.ObjectNumber >= ID_SEARCH_OBJECT1 && item.ObjectNumber <= ID_SEARCH_OBJECT4) && item2.ObjectNumber == ID_EXPLOSION;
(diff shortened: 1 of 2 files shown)


TombEngine/Specific/level.cpp
```diff
@@ -1822,13 +1822,18 @@ void GetCarriedItems()
...
-					abs(item2.Pose.Position.y - item.Pose.Position.y) < CLICK(1) &&
-					Objects[item2.ObjectNumber].isPickup)
+					abs(item2.Pose.Position.y - item.Pose.Position.y) < CLICK(1))
 				{
-					item2.CarriedItem = item.CarriedItem;
-					item.CarriedItem = linkNumber;
-					RemoveDrawnItem(linkNumber);
-					item2.RoomNumber = NO_VALUE;
+					if (Objects[item2.ObjectNumber].isPickup ||
+						(item.ObjectNumber == ID_SEARCH_OBJECT4 && item2.ObjectNumber == ID_EXPLOSION))
```
Lwmte: I am not sure we should have this hack for `ID_EXPLOSION`? TRLE builders are not familiar with search object 4 behaviour, and certainly this seems like a hack, so probably it's better just to omit it.

PR #1891, opened 2026-02-18
BIG_RAT Reworked (tr1_big_rat.cpp)

AlexxxQ:
added the ability to swim in the swamp
added ripple effect on the water
fixed a bug with detecting the enemy's boxnumber in a room with water changed the constants to adapt the rat's behavior to TEN the attack block has been moved to a separate function to remove copy-paste

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,99 +1,85 @@
 name: Bug report
-description: Create a report to help us understand and diagnose your issue. Your contribution is welcomed and valued! It is highly recommended that you are using the latest version before submitting a bug report.
-title: "[ bug Report ]"
-labels: Awaiting Triage
+description: Create a report to help us understand and triage your issue.
+labels: 
+  - Awaiting Triage
 
 body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Bug Report
 
-- type: markdown
-  attributes:
-    value: |
-      Please follow this document in order to report a bug. It is highly recommended that you are using the latest version before submitting a bug report.
+        Please follow this document carefully to report a bug.
 
-- type: dropdown
-  attributes:
-    label: TombEngine version
-    description: |
-      Please select the TombEngine Version from the dropdown list. 
-    options:
-      - Development Build
-      - v1.8.0 (latest version)
-      - v1.7.1
-      - v1.7.0
-      - v1.5
-  validations:
-    required: true
+        > **Important**: It is highly recommended that you use the latest version before submitting a bug report.
 
-- type: dropdown
-  attributes:
-    label: Tomb Editor version
-    description: |
-      Please select the Tomb Editor version used from the dropdown list.
-    options:
-      - Development Build
-      - v1.8.0 (latest version)
-      - v1.7.2
-      - v1.7.1
-      - v1.7.0
-  validations:
-    required: true
+  - type: dropdown
+    attributes:
+      label: Tomb Engine Version
+      description: |
+        Please select the TombEngine version you are using.
+      options:
+        - Development build 
+        - v1.10.1 (latest public release)
+        - v1.10.0  
+    validations:
+      required: true
 
-- type: textarea
-  attributes:
-    label: Describe the bug
-    description: |
-      Please provide A clear and concise description of what the bug is.
-    placeholder: |
-      Your bug report here.
-  validations:
-    required: true
+  - type: checkboxes
+    attributes:
+      label: Development Version
+      description: Are you submitting this report from a development build that has not been officially released?
+      options: 
+        - label: "I am using an unofficial development version."
+        - label: "I am using an official release."
+        - label: "I am using an official pre-release."
+    validations:
+      required: true
 
-- type: textarea
-  attributes:
-    label: To Reproduce
-    description: |
-      To reproduce the behaviour, please provide detailed steps for the development team to follow. This can be done through screenshots or a written guide
+  - type: textarea
+    attributes:
+      label: Describe the Bug
+      description: |
+        Please provide a clear and concise description of what the issue is.
+      placeholder: |
+        Your bug report here.
+    validations:
+      required: true
 
-      **If the bug cannot be reproduced, and if the issue is not adequately explained, it will be closed without further investigation**
-    placeholder: |
-     Provide detailed reproducible steps here. 
-  validations:
-    required: true
+  - type: textarea
+    attributes:
+      label: To Reproduce
+      description: |
+        Please provide detailed steps to reproduce the issue.
+        
+        **Note**: If the bug cannot be reproduced or the issue is not clearly explained, it may be closed without further investigation.
+      placeholder: |
+        Step-by-step reproduction instructions here.
+    validations:
+      required: true
 
-- type: textarea
-  attributes:
-    label: Expected Behaviour
-    description: |
-      A clear and concise description of what you expected to happen.
-    placeholder: |
-      A description of what should happen here.
-  validations:
-    required: true
+  - type: textarea
+    attributes:
+      label: Expected Behaviour
+      description: |
+        What did you expect to happen?
 
-- type: textarea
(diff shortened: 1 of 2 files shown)


TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -6,36 +6,38 @@
...
-	constexpr auto BIG_RAT_LAND_BITE_ATTACK_RANGE  = SQUARE(BLOCK(0.34f));
-	constexpr auto BIG_RAT_POUNCE_ATTACK_RANGE	   = SQUARE(BLOCK(1 / 2.0f));
-	constexpr auto BIG_RAT_WATER_BITE_ATTACK_RANGE = SQUARE(BLOCK(0.3f));
+	constexpr auto BIG_RAT_LAND_BITE_ATTACK_RANGE = SQUARE(BLOCK(0.34f));
+	constexpr auto BIG_RAT_POUNCE_ATTACK_RANGE = SQUARE(BLOCK(1.0f)); // 1.5f in OG, but seems too far for TEN.
 
-	constexpr auto BIG_RAT_REAR_POSE_CHANCE = 1 / 128.0f;
-	constexpr auto BIG_RAT_SWIM_UP_DOWN_SPEED = 32;
-	constexpr auto BIG_RAT_WATER_SURFACE_OFFSET = 10;
+	constexpr auto BIG_RAT_REAR_POSE_CHANCE = 256.0f / 32768.0f; // = 0.0078125 (1/128)
```
Lwmte, line 32: I believe it's not necessary to complicate constants like that. `1 / 128.0f` is more readable than `256.0f / 32768.0f`.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -6,36 +6,38 @@
...
+using namespace TEN::Collision::Room;
 using namespace TEN::Math;
+using namespace TEN::Effects::Ripple;
 
 namespace TEN::Entities::Creatures::TR1
 {
-	constexpr auto BIG_RAT_BITE_ATTACK_DAMAGE	= 20;
-	constexpr auto BIG_RAT_POUNCE_ATTACK_DAMAGE = 25;
+	constexpr auto BIG_RAT_BITE_ATTACK_DAMAGE = 20;
+	constexpr auto BIG_RAT_POUNCE_ATTACK_DAMAGE = 20;
```
Lwmte, line 27: Why damage was reduced here?

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -65,10 +67,6 @@ namespace TEN::Entities::Creatures::TR1
 		BIG_RAT_ANIM_SWIM = 9,
 		BIG_RAT_ANIM_WATER_BITE_ATTACK = 10,
 		BIG_RAT_ANIM_WATER_DEATH = 11,
-
-		// NOTE: These animations don't exist for the TR2 rat. -- TokyoSU 2022.08.10
-		BIG_RAT_ANIM_RUN_FORWARD_TO_SWIM = 12,
```
Lwmte, line 70: Maybe we should keep these constants for forward compatibility.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -77,32 +75,47 @@ namespace TEN::Entities::Creatures::TR1
...
 
-		if (TestEnvironment(ENV_FLAG_WATER, item))
+		if (TestEnvironment(ENV_FLAG_WATER, item) || TestEnvironment(ENV_FLAG_SWAMP, item))
 			SetAnimation(item, BIG_RAT_ANIM_SWIM);
 		else
 			SetAnimation(item, BIG_RAT_ANIM_IDLE);
 	}
 
-	bool RatOnWater(ItemInfo* item)
+	static void PerformAttack(ItemInfo* item, CreatureInfo* creature,
```
Lwmte, line 84: You can keep function declaration in one line, it's not necessary to split it in two lines.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -77,32 +75,47 @@ namespace TEN::Entities::Creatures::TR1
...
-				creature.LOT.Drop = -BLOCK(20);
-			}
-			else
-			{
-				creature.LOT.Step = CLICK(1);
-				creature.LOT.Drop = -CLICK(1.5f);
-			}
+	// Resolves enemy box number when enemy is in a water room below the rat's dry room.
+	// This allows the rat to detect and pathfind toward enemies in water.
+	static void ResolveEnemyBoxNumber(CreatureInfo* creature, AI_INFO& ai)
```
Lwmte, line 99: I don't fully understand the purpose of this function, and why it's exclusive to rat? If it resolves a certain pathfinding culprit for amphibious enemies, probably it's better to make this function generic for all amphibious enemies.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
```diff
@@ -116,31 +129,54 @@ namespace TEN::Entities::Creatures::TR1
...
-			if (doWaterDeath)
+			if (isInWater)
 				CreatureFloat(itemNumber);
 		}
 		else
 		{
 			AI_INFO ai;
 			CreatureAIInfo(item, &ai);
 
+			// Restrict step/drop on land to prevent falling off ledges.
```
Lwmte, line 154: I assume that `RatOnWater` function is unused now and this code replaces it? In any case, I recommend to move LOT modification code to a separate subfunction. You can look at crocodile code for reference (function `SetCrocodileWater`).

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: ok

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: i used as reference tomb1main and TRX

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: > Maybe we should keep these constants for forward compatibility.

Rat doesn't have 12 animation, maybe it was a preparation for CROCORAT creature, but this is not OG behavior.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: ok

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: ResolveEnemyBoxNumber help find boxnumber, if creature lost it. if the rat target (creature->Enemy) does not have a boxnumber defined, the func tries to calculate it from the enemy's(Lara) room sector and immediately updates ai.enemyZone. Gpt chan said it's a "must have" func, but while testing, I didn't see any difference. I tried adding this feature to crocodile today and there is no difference either, it works fine without it, so I will remove ResolveEnemyBoxNumber from rat and crocodile. Designers should know ONE RULE: room with water should have a ramp from the highest ceiling sector to the lowest floor sector of the water room, each sectors of this ramp should be not bigger than CLICK (1.5f) high, otherwise amphibious will not find the boxnumber of Lara.

TombEngine/Objects/TR1/Entity/tr1_big_rat.cpp
AlexxxQ: IsInWater is the same RatOnWater, but calls once and sets a flag in different water cases . I'll take a look at croco

PR #1909, opened 2026-03-03
Fix cobra tr3 (v3)

AlexxxQ:
fixed cobra jerking with partial loss of LOS
(did a little trick: after losing LOS, `EnemyIsVisible` working few more frames.)

added  condition to prevent waking up Cobra, if Player is +-1block higher then Cobra
added  drop speed rotation compensation when Lara is airborne(jump over cobra)
added  4 bones for CreatureJoint rotation
added  enhanced vertical tilt for accurate adjusting CreatureJoint
added  1 more CobraAttackJoint to increase attack chance

### DIFF
--- a/TombEngine/Objects/TR3/Entity/tr3_cobra.cpp
+++ b/TombEngine/Objects/TR3/Entity/tr3_cobra.cpp
@@ -19,7 +19,7 @@ namespace TEN::Entities::Creatures::TR3
 
 	constexpr auto COBRA_ATTACK_RANGE = SQUARE(BLOCK(1));
 	constexpr auto COBRA_AWARE_RANGE = SQUARE(BLOCK(1.25f));
-	constexpr auto COBRA_VERTICAL_AWARE_RANGE = SQUARE(BLOCK(0.5f));
+	constexpr auto COBRA_VERTICAL_AWARE_RANGE = BLOCK(0.5f);
 	constexpr auto COBRA_SLEEP_RANGE = SQUARE(BLOCK(2.5f));
 
 	constexpr auto COBRA_DISTURBANCE_VELOCITY = 15.0f;


TombEngine/Objects/TR3/Entity/tr3_cobra.cpp
```diff
@@ -143,7 +167,9 @@ namespace TEN::Entities::Creatures::TR3
 					item->HitPoints = NOT_TARGETABLE;
 				}
 
-				if (AI.distance < COBRA_AWARE_RANGE && creature->Enemy && creature->Enemy->HitPoints > 0)
+				
+				if (AI.distance < COBRA_AWARE_RANGE && creature->Enemy && creature->Enemy->HitPoints > 0 &&
+					abs(AI.verticalDistance) < COBRA_VERTICAL_AWARE_RANGE && !creature->Enemy->Animation.IsAirborne) // Wake up if Lara grabs the edge in AWARE_RANGE.
```
Lwmte, line 172: I believe this additional condition will now work in all cases, so cobra will never wake up when Lara is airborne (jumping, falling etc), and additionally it will never wake up if vertical distance is more than 0.5 blocks. While we can tolerate `IsAirborne`, not sure that ignoring enemy above/below just 0.5 blocks is correct.

AlexxxQ: 1. Cobra can't do damage higher then 2.5 clicks(560 units), that's why i set limit to 0.5 blocks. You can change to 0.75, but then cobra will become an easy, helpless target for player. 

2. Without `IsAirborne` you can trigger cobra from safe place, look at the attached video

https://github.com/user-attachments/assets/c60b31f6-123f-436f-bca1-67ea2f626195

As OG have no such cases with placed cobra on high blocks, we can freely interpret her behavior based on the builders wish.

Lwmte: Let's say I want a scenario when Lara has to fall through a pit and there's a snake in the middle:

Image

Is it possible to achieve with new conditions?

PR #1912, opened 2026-03-06
Fix for flying creatures (PredictTargetPosition Issue)

AlexxxQ:
The issue is in CreatureMood::Attack. When PredictTargetPosition() is called, it uses LOT->Target (from the previous frame) as the smoothing base for interpolation. But LOT->Target.y from the last frame already had Y1 added to it. Then Y1 gets added again via LOT->Target.y += bounds.Y1, causing a double-application that accumulates each frame and pushes the target Y higher and higher — making flying creatures hover one block (or more) above Lara instead of at her level. In 1.10, the code was simply LOT->Target = enemy->Pose.Position, setting Y directly without any smoothing interference, then adding Y1 once. The fix is to override the Y directly instead of doing += bounds.Y1 through a smoothed position/

### DIFF
--- a/TombEngine/Game/control/box.cpp
+++ b/TombEngine/Game/control/box.cpp
@@ -2643,8 +2643,12 @@ void CreatureMood(ItemInfo* item, AI_INFO* AI, bool isViolent)
 
 	case MoodType::Attack:
 	{
-	// Flying creatures target enemy's upper body when on land.
-	auto targetOffset = (LOT->Zone == ZoneType::Flyer && Lara.Control.WaterStatus == WaterStatus::Dry)
+	// Flying creatures target enemy's upper body when enemy is not in water.
+	bool isEnemyOnLand = enemy->IsLara()
+		? (GetLaraInfo(*enemy).Control.WaterStatus == WaterStatus::Dry)		 // Lara.
+		: !TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, enemy->RoomNumber); // Other Creatures.
+
+	auto targetOffset = (LOT->Zone == ZoneType::Flyer && isEnemyOnLand)
 		? Vector3i(0, GetClosestKeyframe(*enemy).BoundingBox.Y1, 0) : Vector3i(0, 0, 0);
 
 	LOT->Target = PredictTargetPosition(*item, *enemy, targetOffset);
--- a/TombEngine/Game/control/box.cpp
+++ b/TombEngine/Game/control/box.cpp
@@ -2644,10 +2644,10 @@ void CreatureMood(ItemInfo* item, AI_INFO* AI, bool isViolent)
 	case MoodType::Attack:
 	{
 	// Flying creatures target enemy's upper body when on land.
-	int headOffset = (LOT->Zone == ZoneType::Flyer && Lara.Control.WaterStatus == WaterStatus::Dry)
-		? GetClosestKeyframe(*enemy).BoundingBox.Y1 : 0;
+	auto targetOffset = (LOT->Zone == ZoneType::Flyer && Lara.Control.WaterStatus == WaterStatus::Dry)
+		? Vector3i(0, GetClosestKeyframe(*enemy).BoundingBox.Y1, 0) : Vector3i(0, 0, 0);
 
-	LOT->Target = PredictTargetPosition(*item, *enemy, headOffset);
+	LOT->Target = PredictTargetPosition(*item, *enemy, targetOffset);
 	LOT->RequiredBox = enemy->BoxNumber;
 
 	break;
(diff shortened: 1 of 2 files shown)


TombEngine/Game/control/box.cpp
```diff
@@ -2650,7 +2650,7 @@ void CreatureMood(ItemInfo* item, AI_INFO* AI, bool isViolent)
 		if (LOT->Zone == ZoneType::Flyer && Lara.Control.WaterStatus == WaterStatus::Dry)
 		{
 			auto& bounds = GetClosestKeyframe(*enemy).BoundingBox;
-			LOT->Target.y += bounds.Y1;
+			LOT->Target.y = enemy->Pose.Position.y + bounds.Y1;
```
Lwmte: I checked with git blame, and this code was not `LOT->Target = enemy->Pose.Position + bounds.Y1`. Previous change to this line was 4 years ago, which means that either this problem existed for last 4 years, or manifests because of other reasons.

Additionally, this code is similar to OG code in tomb4: https://github.com/Trxyebeep/tomb4-legacy/blob/a0620f455350aa4947bbcbe18b417b0368e69fad/TOMB4/game/box.cpp#L661

Therefore, additional research must be done to understand real reason for this issue.

Lwmte (changes requested):

TombEngine/Game/control/box.cpp
AlexxxQ: MASTER BRANCH & TOMB4:          
  `LOT->Target = enemy->Pose.Position; LOT->Target.y += bounds.Y1;`
DEV BRANCH: 
 `LOT->Target = PredictTargetPosition(*item, *enemy);  LOT->Target.y += bounds.Y1;`

TombEngine/Game/control/box.cpp
```diff
@@ -2899,7 +2897,7 @@ void GetCreatureMood(ItemInfo* item, AI_INFO* AI, bool isViolent)
 	}
 }
 
-Vector3i PredictTargetPosition(ItemInfo& sourceItem, ItemInfo& targetItem)
+Vector3i PredictTargetPosition(ItemInfo& sourceItem, ItemInfo& targetItem, int targetYOffset)
```
Lwmte: Probably good idea is to provide `Vector3i` offset for any future cases when we may need also horizontal offsetting.

TombEngine/Game/control/box.cpp
```diff
@@ -2642,18 +2642,16 @@ void CreatureMood(ItemInfo* item, AI_INFO* AI, bool isViolent)
 		break;
 
 	case MoodType::Attack:
-		// ATTACK: Go directly to enemy's position.
-		LOT->Target = PredictTargetPosition(*item, *enemy);
-		LOT->RequiredBox = enemy->BoxNumber;
+	{
+	// Flying creatures target enemy's upper body when on land.
+	int headOffset = (LOT->Zone == ZoneType::Flyer && Lara.Control.WaterStatus == WaterStatus::Dry)
```
Lwmte: I believe we must also fix this problem with `CreatureMood` directly referencing `Lara` global instead of enemy. If creature's `enemy` is `Lara`, we can use this code path with `WaterStatus`, but if enemy isnt Lara, we must fallback to zero offset in any case.

TombEngine/Game/control/box.cpp
AlexxxQ: done

TombEngine/Game/control/box.cpp
AlexxxQ: Why you need it? This offset is the upper point of the BoundingBox of any creature, it's not like +1 block above the head

Lwmte: I mean, `Lara.Control.WaterStatus == WaterStatus::Dry` condition will fire regardless who is the enemy - let's say, harpy attacks wild boar, while Lara is in water - it will make harpy attack at wild boar's base position. This will be solved by not referencing `Lara` directly, but by checking `enemy.IsLara()` and then getting `GetLaraInfo()` from the `enemy` and accessing water status from it.

We also generally try to avoid using `Lara` and `LaraItem` globals in scenarios which imply that not only Lara may be an enemy. It's not necessary to bulldoze every enemy's code for this change, but this generic AI function may worth fixing.

TombEngine/Game/control/box.cpp
Lwmte: Sorry, I misunderstood, thank you.

TombEngine/Game/control/box.cpp
AlexxxQ: done. added `isEnemyOnLand` var, in which Lara is checked by `WaterStatus`, other enemies by `TestEnvironment`

PR #1913, opened 2026-03-06
Fix blood effect for spiky traps

AlexxxQ:
There is no blood effect when taking damage from a spiky wall\ceiling. The reason is blood spawns always underground.
Solution:
invert the math symbols in SpikyCeiling.cpp line 96
and SpikyWall.cpp line 146


Lwmte (commented):
I see no issues with this PR.

PR #1915, opened 2026-03-07
Fix winged mutant fly turning and target offset for some creatures

AlexxxQ:
The Winged mutant has 3 bugs: 
1. No MaxTurn in FLY Case. 
2. Wrong TargetOffset formula for shooting attack.
3.  BodyPartExplode function do infinite loop if the body parts touch the water, it sometimes freeze or crash engine.

Wrong TargetOffset formula is epic bug, because it corrupted targeting at Lara in water rooms. These creatures use it: Centaur(TR1), Natla(TR1), WingedMutant(TR1), KnifeThrower(TR2), ScubaDiver(TR3)

### DIFF
--- a/TombEngine/Game/missile.cpp
+++ b/TombEngine/Game/missile.cpp
@@ -44,10 +44,8 @@ void ShootAtEnemy(Vector3i target, ItemInfo* item, int index)
 		target = item->Pose.Position;
 
 	auto bounds = GameBoundingBox(item);
-	auto targetWithOffset = Vector3(
-		target.x,
-		item->Pose.Position.y + bounds.Y2 + (bounds.Y1 - bounds.Y2) * 0.75f,
-		target.z);
+	auto yOffset = bounds.Y2 - (bounds.GetHeight() * 0.75f);
+	auto targetWithOffset = Vector3(target.x, item->Pose.Position.y + yOffset, target.z);
 
 	// Apply slight random scatter.
 	auto randomOrient = EulerAngles(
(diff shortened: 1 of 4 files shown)


TombEngine/Game/missile.cpp
```diff
@@ -43,9 +43,10 @@ void ShootAtEnemy(Vector3i target, ItemInfo* item, int index)
 	if (Vector3i::Distance(item->Pose.Position, target) > TARGET_DEVIATION_THRESHOLD)
 		target = item->Pose.Position;
 
+	auto bounds = GameBoundingBox(item);
 	auto targetWithOffset = Vector3(
 		target.x,
-		target.y - (GameBoundingBox(item).GetHeight() * 0.75f),
+		item->Pose.Position.y + bounds.Y2 + (bounds.Y1 - bounds.Y2) * 0.75f,
```
Lwmte: I don't understand this formula. Why you are using `bounds.Y1 - bounds.Y2` instead of `bounds.GetHeight()`?

AlexxxQ: `item->Pose.Position.y + bounds.Y2 + (bounds.Y1 - bounds.Y2) * 0.75f` is OG formula from TR1, TR3.  
Old formula: 
Lara on Land - `target.y` is under Laras legs+offset = target to body.
Lara in water - `target.y` is `WaterHeight` (~at Laras head)+ offset = target above head, (~512 units upper then head!)

Lwmte: I understand, I mean technically `bounds.Y1 - bounds.Y2` should be equal to `bounds.GetHeight()`. Can you please check if that's true or not in your case?

AlexxxQ: `GetHeight()`  use `abs(Y2 - Y1);`   Anyway i tested `item->Pose.Position.y + bounds.Y2 - bounds.GetHeight() * 0.75f` and it JUST WORKS!

PR #1926, opened 2026-03-17
Fix vertex move effect speed in water rooms while turning camera on x axis

AlexxxQ:
To test delete BIN folder in the shaders and check room.fx gbuffer.fx is updated
The vertex movement phase was considered like this:

clipPos = screenPos.xy / screenPos.w
factor = Frame + clipPos.x * 320

clipPos.x is the on-screen X-coordinate, which changes extremely when the camera
is rotated left/right.
Because of this, the sine/cosine for vertex displacement received an additional
phase acceleration from the camera, and visually the move/reflection effect
went faster (as if by ~2 times).


Lwmte:
I did a visual comparison of the effect in `develop` and this branch, and I must say that this PR removes the distortion effect altogether rather than fixing it. Consider changing the shader formula to preserve the effect but make it less dependent on a camera angle.

AlexxxQ:
it works in 1.10 and stop working in 1.11. I will find the reason/


https://github.com/user-attachments/assets/9cce713c-45d5-4915-90e3-0aae334ca52b

AlexxxQ:
fixed. To test  you need to manually change the rooms.hlsl and gbuffer.hlsl files in the shader folder, cause they doesnt compile, if test level already exist.

Stranger1992:
@AlexxxQ Ive just tested this, and I think the effect is now fixed. Are you happy to merge this?

Stranger1992 (approved):
Using a direct comparison between `develop` and this branch, the broken effect is now fixed. Other devs have also expressed satisfaction with the implementation and resulting effect.

PR #1939, opened 2026-03-25
Add ripples bubbles effect to falling crumbling platform

AlexxxQ:
https://github.com/user-attachments/assets/91e2b07e-f191-4cc6-a72b-6a4c29350dc5


TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -181,7 +188,36 @@ namespace TEN::Entities::Traps
 			// Update room number.
 			int probedRoomNumber = pointColl.GetRoomNumber();
 			if (item.RoomNumber != probedRoomNumber)
+			{
+				// Spawn 6 splashes at random positions across platform when entering water.
```
Lwmte: Probably better idea is to calculate positions using `GetJointPositon`? It's a subtle improvement, but maybe it will look more believable.

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -181,7 +188,36 @@ namespace TEN::Entities::Traps
...
+					}
+				}
+
 				ItemNewRoom(itemNumber, probedRoomNumber);
+			}
+
+			// Spawn bubbles every frame while sinking underwater.
+			if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, item.RoomNumber))
+			{
+				for (int i = 0; i < 6; i++)
```
Lwmte: Please demagic `6` and enclose for loop body into curly brackets `{}`.

AlexxxQ: done

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -181,7 +192,38 @@ namespace TEN::Entities::Traps
...
 			if (item.RoomNumber != probedRoomNumber)
+			{
+				// Spawn splash for each bone of the platform when entering water.
+				if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, probedRoomNumber) &&
+					!TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, item.RoomNumber))
+				{
+					int waterHeight = GetPointCollision(item.Pose.Position, probedRoomNumber).GetWaterTopHeight();
+
+					// Force-update bone matrices for this logic frame so sphere positions are current.
+					g_Renderer.UpdateItemAnimations(itemNumber, true);
```
Lwmte: I don't fully understand why are you forcing update of bone matrices. `item.GetSpheres` implicitly calls `UpdateItemAnimations`, if it wasn't called for the current game frame (see RendererHelper.cpp, line 431).

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -181,7 +192,38 @@ namespace TEN::Entities::Traps
...
+				}
+
 				ItemNewRoom(itemNumber, probedRoomNumber);
+			}
+
+			// Spawn bubbles every frame while sinking underwater.
+			if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, item.RoomNumber))
+			{
+				constexpr auto bubbleCount = 6;
+				for (int i = 0; i < bubbleCount; i++)
```
Lwmte: Bubble spawn can get costly if you have several platforms sinking at the same time. Consider adding `Random::TestProbability` method to skip bubble generation in certain percentage of cases.

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
AlexxxQ: right, removed

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
AlexxxQ: done

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -180,8 +189,39 @@ namespace TEN::Entities::Traps
...
+
 				ItemNewRoom(itemNumber, probedRoomNumber);
+			}
+
+			// Spawn bubbles every frame while sinking underwater.
+			if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, item.RoomNumber))
+			{
+				for (const auto& sphere : spheres)
+				{
+					if (Random::TestProbability(1 / 2.0f))
```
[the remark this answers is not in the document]
AlexxxQ: Done. Tuned underwater bubble spawning to be depth-aware: 50% chance for the first block below the water surface, then 10% deeper down. This reduced the number of bubbles twice without any visual difference.

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -180,8 +189,39 @@ namespace TEN::Entities::Traps
 
 			// Update room number.
 			int probedRoomNumber = pointColl.GetRoomNumber();
+			auto spheres = item.GetSpheres();
+
```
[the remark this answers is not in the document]
AlexxxQ: Done. Moved GetSpheres() into water-only branches so falling platforms no longer fetch sphere data every frame on dry land.

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -180,8 +189,39 @@ namespace TEN::Entities::Traps
...
+				{
+					int waterHeight = GetPointCollision(item.Pose.Position, probedRoomNumber).GetWaterTopHeight();
+
+					for (const auto& sphere : spheres)
+					{
+						SplashSetup.Position = Vector3(sphere.Center.x, (float)(waterHeight - 1), sphere.Center.z);
+						SplashSetup.SplashPower = GenerateFloat(fallVel * 0.5f, fallVel * 2.0f);
+						SplashSetup.InnerRadius = sphere.Radius;
+						SetupSplash(&SplashSetup, probedRoomNumber);
+					}
```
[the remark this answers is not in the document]
AlexxxQ: Done. Added a safeguard cooldown for splash sound and drip spawning to prevent stacked audio and excessive drips from rapid repeated splash calls.

TombEngine/Game/effects/Splash.h
```diff
@@ -2,7 +2,7 @@
 
 namespace TEN::Effects::Splash
 {
-	constexpr auto SPLASH_EFFECT_COUNT_MAX = 8;
+	constexpr auto SPLASH_EFFECT_COUNT_MAX = 64;
```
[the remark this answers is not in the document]
AlexxxQ: Done. Reduce SPLASH_EFFECT_COUNT_MAX to 48 and  reduced setupCountMax to 2 for crumbling-platform splash calls only,  now the peak maximum value is 36 splashes at the same time(which is half as much as before) and 2 splashes max per frame for each sphere(very cheap).

TombEngine/Objects/Generic/Traps/crumblingPlatform.cpp
```diff
@@ -180,8 +189,39 @@ namespace TEN::Entities::Traps
...
+			// Spawn bubbles every frame while sinking underwater.
+			if (TestEnvironment(RoomEnvFlags::ENV_FLAG_WATER, item.RoomNumber))
+			{
+				for (const auto& sphere : spheres)
+				{
+					if (Random::TestProbability(1 / 2.0f))
+						SpawnBubble(
+							GeneratePointInSphere(sphere),
+							item.RoomNumber, GenerateInt(32, 256), GenerateInt(BLOCK(0.1f), BLOCK(0.25f)));
+				}
```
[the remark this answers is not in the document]
AlexxxQ: Done. Tuned underwater bubble spawning to be depth-aware: 50% chance within the first block below the water surface, then 10% deeper down. Its very cheap now.

AlexxxQ:
All comments addressed, ready for testing.

Christian-Croft:
**Quick Test**

**Overall**:
- ✅ Produces TEN splash effect when on directly on top of the water surface (small splash)
Add a TR3/4 splash effect when above the water surface with more or less intensity depending of the height.
As expected

- ❌ **Note**: The splash effect seems to be oriented incorrectly as it looks "inward" - whereas in classic Tomb Raider the effect is straighter and slightly outward as seen on that [video](https://youtu.be/UdDLNvHtx5A?si=7Yozb0F90cy51gFV&t=326) 
Here how it's currently looking in TEN
Capture d’écran 2026-05-14 à 09 54 54

**Water Room:**
- ✅ Tested on different Water Room:
Classic water room: ok
flipped dry to water room: ok 
flipped with SetFlag() function: ok
Water room set in different height in the same main room; resulting in the correct effect each time.

- ❌ On shallow water (less than 352 units deep), the platform must be at least 7 full clics above the water surface to produce the effect - if bellow that height, no effect is produced when falling into the water.
 ✅ If the water is 352 units deep or more, the effect is produced even if the crumbling platform is right on the water surface.

**Objects:**
- ✅ Tested with "classic" crumbling platforms from TEN website: Result are as expected
- ❌**Note**: @Stranger1992 TR1 style crumbling platform from TEN website seems to have huge sphere collision radius, resulting in a bad result since that effect relies on sphere size - The asset should be updated

- ✅ Tested with "custom" crumbling platform (modified TR1) with spread out meshes: Results are as expected where the effect is done on each individual mesh depending on their position

- ✅ Tested with "custom" crumbling platform (modified TR1) with 1 mesh: Results are as expected where the effect is applied only on the existing mesh 

**Performances**: (Tested on a Virtual machine running W11 on macOS Tahoe - must be tested on a proper Windows machine)
- ✅ No performance issue or frame drop with few crumbling platform

- ✅❌ **?** Extreme Test with 60 crumbling platforms: Do not produce the effect for all the objects like shown in this video. Do the effect is meant to be stoped if called a certain amount of time ? It looks like the effect stops to be called after 6 platforms at the same time.
https://github.com/user-attachments/assets/df6143e5-17fd-4c02-946e-58cd88f9c7ec


Tell me if other tests are needed.

AlexxxQ:
@Christian-Croft 

Thanks for the test!
1. I checked the OG TR5 splashes and you are right - the inner radius in TEN is reversed, so I fixed it:

https://github.com/user-attachments/assets/a82d969b-f14a-48f1-872b-8b805bc2d5ee

2.  I can't reproduce the shallow water bug, but its absolutely true that the vertical splash effect depends on the platform's falling speed velocity; if the speed velocity is close to zero, the effect of vertical splashes is not noticeable.

https://github.com/user-attachments/assets/af3ac395-5efe-4717-af5a-f50398a49cdd

3.  All 7 spheres of platform must be recalculated.

4.  "Extreme Test with 60 crumbling platforms" is really extreme! Every splashes also spawn bubbles, drips and play the splash SFX once per sphere, so it's x6-factor x60 platforms! 
 Lwmte and Copilot reviews asked to optimize it,  so I did special debug profiling for the platform to optimize the number of splash slots without visual loss of quality. Tests showed that 48 effect slots (at the same time) are enough to visualize successively falling platforms beneath a running Lara. This is very cheap now, but after property system implementation it can be increased.

https://github.com/user-attachments/assets/5759fb4c-89da-4380-a830-fd8a7bda136c

Lwmte:
I am adding a fallback radius for cases when sphere radius is significantly bigger than overall object's bounding box extents. In such cases, engine will use average bounding box size for splash radius instead of spheres. This way the feature will be compatible with legacy assets ported to TEN as is.

Lwmte (approved):

TombEngine/Game/effects/Splash.cpp
```diff
@@ -83,21 +85,28 @@ namespace TEN::Effects::Splash
 				splashSetupCount++;
 			}
 
-			if (splashSetupCount == SETUP_COUNT_MAX)
+			if (splashSetupCount == setupCountMax)
 				break;
 		}
 
-		SpawnSplashDrips(Vector3(setup->Position.x, setup->Position.y - 15, setup->Position.z), room, 32);
+		if (SplashTimeout == 0)
```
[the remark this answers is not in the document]
Lwmte: AI glitch.

PR #1966, opened 2026-04-16
Fix: Consistent initial state for paused flag in Timer.Create

davidmarr:
## Checklist

- [x] I have added a changelog entry to the `CHANGELOG.md` file on the branch/fork (if it is an internal change, this is not needed).
- [x] This pull request meets the Coding Conventions standards: https://github.com/MontyTRC89/TombEngine/blob/master/CONTRIBUTING.md#coding-conventions

N/A

## Pull request description
`Timer.Create()` initialized `thisTimer.paused = true` while thisTimer.active = false. This caused IsPaused() to return true for a timer that was never started — a semantically incorrect result. A recently added fix to `Timer:Stop()` (which correctly sets `paused = false`) exposed an internal inconsistency: two logically equivalent states ("not running") had different paused values.

|         State         | active | paused (before) | paused (after) |
|:---------------------:|:------:|-----------------|----------------|
| After `Create()`        |  false | true ← wrong    | false ← fixed  |
| After `Start()`         |  true  | false           | false          |
| After `SetPaused(true)` |  true  | true            | true           |
| After `Stop()`          | false  | false           | false          |

**Why** `paused = true` on creation was wron

1. Inconsistency with `Stop()`: Both `Create` and `Stop` produce a "not running" timer. `Stop()` correctly sets paused = false, so `Create` setting `paused = true` was contradictory. If "created = paused" were intentional, then Stop() would have a bug — but Stop() is correct. 
2. Semantic error: "Paused" means "was running, then suspended." A timer that was never started is idle, not paused. No standard timer API (JavaScript, C#, Unity, Python) considers a newly constructed timer to be in a paused state.
3. False positives in Node Editor: The "If timer is paused..." conditional node returned true for timers that were created but never started, misleading level builders who use this check without also checking "If timer active...".

**Backward compatibility analysis**

Every code path reading `paused` has been audited:

- `Decrease`: Guarded by `if t.active and not t.paused` — with `active = false,` `paused` is never evaluated. No effect.
- `UpdateAll` (display + color): Inside `if t.active` — unreachable when `active = false`. No effect.
- `IsTicking()`: Returns `not paused and active`. With `active = false`, result is `false` regardless of `paused`. No effect.
- `StartTimer` node: Checks `IsPaused() and IsActive()`. With `active = false`, short-circuits to `false` either way. No effect.
- `Timer:Start()`: Explicitly sets paused = false. No effect.
- `Timer:SetPaused(p)`: Overwrites paused. No effect.
- `EventSequence:IsPaused()`: Internal timers (__TEN_*) are started immediately after creation. No effect.

The only observable change is that `IsPaused()` now correctly returns `false` for a timer that exists but was never started.

**Changes**

`Timer.Create`:
```diff
- thisTimer.paused = true
+ thisTimer.paused = false
```
`@treturn` documentation:
```diff
- -- @treturn[1] Timer The timer in its paused state
+ -- @treturn[1] Timer The timer in its idle state, not yet started
```
No runtime behavior changes for any existing code. No breaking changes.


Lwmte:
This pull request does not take into consideration cases when builder may have used `Timer:IsPaused()` method or "If timer is paused..." conditional node for created but inactive timers. While it may be more semantically correct, it breaks compatibility with projects potentially using this workaround. Therefore I can't approve this pull request.

PR #1972, opened 2026-04-24
Correct handling of strings

MontyTRC89:
This pull request solves once for all the problems with special characters in strings. Now every string in TEN is a std::string and we use std::wstring only internally and the conversion is done in a cross platform way.


Stranger1992 (approved):
## Display Strings

**Using `Noto Sans` spritefont file , generated with all characters**

Image

* I tried using Arabic ( `من فضلك جرب شغلي كمان` ) as a display string and it didnt work. This may be down to D-DIN not supporting Arabic characters. But it recognised that there was characters missing, so thats fine with me.
* I tried using Russian `Пожалуйста, проверьте и мою работу!` and displayed
* I tried using Greek `Παρακαλώ δοκιμάστε και τη δουλειά μου!` and it displayed

Image

Image

## File Paths

I renamed my `Screens` folder to `Παρακαλώ δοκιμάστε και τη δουλειά μου` and adjusted the level script accordingly and this worked. It found the image and displayed it.

Approved.

PR #1973, opened 2026-04-24
Cross platform I/O

MontyTRC89:
This pull requests removes Win32 specific code for I/O and uses instead std stream classes for reading, writing and paths handling


TrainWrack (approved):
Tested the branch as per Monty's instruction.

Changed all \\ in gameflow to // and it works without any error. Levels load, scripts load, soundtrack played.

Load / Save worked in big levels without any issue.

PR #1980, opened 2026-04-26
Fix tr5 bats emitter issues & enhanced with tr3 bats effect(v2)

AlexxxQ:
https://github.com/user-attachments/assets/a3f3899f-e343-4875-a89a-ab8b511f5eb8

This PR extends BATS_EMITTER (slot 132) with a second behavior mode based on OCB value:

OCB = 0: TR3-style one-shot flyaway bats.
OCB > 0: existing TEN/TR5 attacking bats behavior is preserved.

What changed:

Added a separate TR3 flyaway path for bat emitter initialization, triggering and update logic.
Kept TR5 attacking bats as a separate path selected through a mode switch.
Added a dedicated Tr3Bats pool so TR3 and TR5 behaviors do not interfere with each other.
Tuned TR3 bat movement to better match original TR3 speed and lifetime.
Fixed TR3 bat orientation so they face their movement direction correctly.
Increased TR3 bat visual scale to better match the original appearance.
Renamed TR5-only constants to explicit TR5_* names for clarity.
Updated rendering so both TR5 bats and TR3 bats are drawn correctly.
Added a one-time random animation phase offset on emitter activation.

Files touched:
TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.h
TombEngine/Renderer/RendererDraw.cpp
TombEngine/Game/Savegame.cpp


Lwmte:
Probably it can be better to create a separate slot ID for TR3 bats? Currently we are in the process of splitting all legacy slots without shady OCB actions.

AlexxxQ:
> Probably it can be better to create a separate slot ID for TR3 bats? Currently we are in the process of splitting all legacy slots without shady OCB actions.

A separate ID slot for OG TR3_BAT_EMITTER? I can do it, but then builders won't be able to customize it, it's a completely hardcoded effect.  The animation and the 32x32 sprite is hardcoded by vertex.UV to the generated triangles.

https://github.com/user-attachments/assets/23a6bdf3-098f-447e-b11b-1df9855e8f86

Lwmte:
Can we just use modern rendering technique but use another slot for it?
As for customization, I'd postpone adding any new OCBs because soon we will have more user-friendly property system in place.

PR #1990, opened 2026-05-06
Apply C++ standard compliance for cross-platform G++ build

MontyTRC89:
This PR makes the code G++ compatible removing all hacks that MSVC compiler accepted but that were out of C++ standard.


TombEngine/Game/Lara/lara_struct.h
```diff
@@ -1026,9 +1026,9 @@ struct Ammo
 	bool		 IsInfinite = false;
 
 public:
-	static unsigned int Clamp(long value)
+	static unsigned int Clamp(long long value)
```
Lwmte: Not sure why `long long` type was enforced here. Ammo counter may never reach such large values.

TombEngine/Game/Lara/lara_flare.cpp
```diff
@@ -75,6 +75,9 @@ void FlareControl(short itemNumber)
 
 	const auto& settings = g_GameFlow->GetSettings()->Flare;
 
+	if (!flareItem.Data.is<int>())
```
Lwmte: I believe it's better to do an early `return` in the beginning of the function, if flareItem data is not int instead of randomly recreating it here in the middle of the control loop, because flare control is never meant to be called on an object without this ItemData type.

Lwmte (approved):

PR #1999, opened 2026-05-11
Add (140)BATS_EMITTER_TR3 entity

AlexxxQ:
## Summary

Adds the TR3 bat swarm emitter as `TR3_BATS_EMITTER` in object slot 198.

This ports the original TR3-style bat emitter behavior into TEN, including bat spawning, movement, wing animation, interpolation, room ambient lighting, and rendering through the sprite system.

## Details

- Adds `ID_BATS_EMITTER_TR3 = 140`.
- Registers `BATS_EMITTER_TR3` in TEN object IDs and enum generation data.
- Adds `tr3_bats_emitter.cpp/.h`.
- Hooks the emitter into TR3 object initialization and the main update loop.
- Renders the bats as three animated sprite triangles, matching the original TR3 emitter layout.
- Adds a custom UV path for 3D sprites so TR3 bat triangles can use the original per-triangle UV orientation without affecting normal sprite rendering.
- Keeps the custom UV behavior opt-in through `UseCustomUV`, so existing sprite emitters continue using the default mapping.

- Adds savegame support for BATS_EMITTER_TR3:

  Serializes active TR3 bat swarm state into savegames.
  Restores TR3 bat swarm state when loading a save.
  Adds tr3_bats to the savegame FlatBuffers schema as a SwarmObjectInfo vector.
  Regenerates ten_savegame_generated.h from the updated schema.

https://github.com/user-attachments/assets/74010eef-b4ec-4f52-b5b6-8b45c4af7ee0


Lwmte:
The used approach intervenes with renderer too much in my opinion. What I suggest is to take TR5 bats and just use TR3 control routines for it (via separate slot), without changing the rendering.

PR #2002, opened 2026-05-12
Fix TR5 Bats Emitter issues

AlexxxQ:
## Fixes

- Fixed bat target offset storage by changing `XTarget` and `ZTarget` from `byte` to `short`.
  (XTarget and ZTarget must use a signed integer type, because their randomized offsets can be negative)
- Fixed bat steering by using `Geometry::GetShortestAngle()` for angle delta calculation.
- Fixed turn behavior so bats can correctly steer in both directions.
- Replaced raw memory clearing of the TR5 bats array with value initialization so `BatData` are initialized correctly.

## Result

These changes restore more reliable bat movement and targeting behavior and prevent incorrect offset wrapping caused by unsigned target fields.

https://github.com/user-attachments/assets/910d99dd-6e43-429c-ad59-8b17e87b7c40

### DIFF
--- a/TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
+++ b/TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
@@ -207,9 +207,7 @@ void UpdateBats()
 		if (ItemNearTarget(bat->Pose.Position, LaraItem, CLICK(1)))
 		{
 			TriggerBlood(bat->Pose.Position.x, bat->Pose.Position.y, bat->Pose.Position.z, 2 * GetRandomControl(), 2);
-
-			if (LaraItem->HitPoints > 0)
-				DoDamage(LaraItem, BAT_LARA_DAMAGE);
+			DoDamage(LaraItem, BAT_LARA_DAMAGE);
 		}
 
 		Matrix translation = Matrix::CreateTranslation(bat->Pose.Position.x, bat->Pose.Position.y, bat->Pose.Position.z);
--- a/TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
+++ b/TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
@@ -35,7 +35,10 @@ void InitializeLittleBats(short itemNumber)
 		item->Pose.Position.x += CLICK(2);
 
 	if (Objects[ID_BATS_EMITTER].loaded)
-		ZeroMemory(Bats, NUM_BATS * sizeof(BatData));
+	{
+		for (auto& bat : Bats)
+			bat = {};
+	}
 
 	//LOWORD(item) = sub_402F27(ebx0, Bats, 0, 1920);
 }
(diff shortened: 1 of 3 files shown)


MontyTRC89 (approved):

TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
```diff
@@ -201,10 +204,12 @@ void UpdateBats()
 		bat->Pose.Position.y += bat->Velocity * phd_sin(-bat->Pose.Orientation.x);
 		bat->Pose.Position.z += sp * phd_cos(bat->Pose.Orientation.y);
 
-		if (!(i % 1) && ItemNearTarget(bat->Pose.Position, LaraItem, CLICK(1)))
+		if (ItemNearTarget(bat->Pose.Position, LaraItem, CLICK(1)))
 		{
 			TriggerBlood(bat->Pose.Position.x, bat->Pose.Position.y, bat->Pose.Position.z, 2 * GetRandomControl(), 2);
-			DoDamage(LaraItem, BAT_LARA_DAMAGE);
+
+			if (LaraItem->HitPoints > 0)
```
Lwmte: This is not necessary. DoDamage itself does the hit points check and does early exit when hit points are 0 or below 0.

AlexxxQ: ok! removed!

TombEngine/Objects/TR5/Emitter/tr5_bats_emitter.cpp
```diff
@@ -201,7 +204,7 @@ void UpdateBats()
 		bat->Pose.Position.y += bat->Velocity * phd_sin(-bat->Pose.Orientation.x);
 		bat->Pose.Position.z += sp * phd_cos(bat->Pose.Orientation.y);
 
-		if (!(i % 1) && ItemNearTarget(bat->Pose.Position, LaraItem, CLICK(1)))
```
Lwmte, line 204: Why `!(i % 1)` condition was removed? It is still present in original tomb5 codebase: https://github.com/Trxyebeep/TOMB5/blob/master/TOMB5/game/bat.cpp#L223
To me, it looks like this condition makes only every second bat hurt Lara. Removing this condition effectively doubles the damage caused by bat swarm.

AlexxxQ: This condition check index of bats. it was removed, cause `!(i % 1)` always return true and do nothing. It means every bat do damage. 
In OG it was `!(i & 1)` = every second bat do damage, but in TEN `ItemNearTarget` increased damage zone too much and there is almost no difference with  `!(i & 1)` or without it. 
OG used simple bbox logic for damage, so its big difference in damage between OG and TEN.

Difference In practice for x16 bats:
```
!(i % 1) + ItemNearTarget ≈ 100% damage (TEN)
!(i & 1) + ItemNearTarget ≈  95% damage (TEN)
!(i & 1) + bbox           ≈  25% damage (OG)
```
What should I do?  restore OG logic with bbox? 
or increase the divider to imitate OG damage in TEN, for example `!(i % 4)`?

Lwmte: I believe that probably better solution is to determine chance of bat hurting player by using `Random::TestProbability` method, It will look more realistic.

AlexxxQ: Done!

Lwmte (approved):

PR #2005, opened 2026-05-16
Fix TR3 Monkey issues

AlexxxQ:
This PR restores and improves TR3 monkey pickup behavior in TEN.

The monkey now correctly targets and picks up TR3-style pickup items:

- In normal mode, the monkey searches for and picks up SMALLMEDI_ITEM.
- In AI_MODIFY mode, the monkey searches for and picks up KEY_ITEM4.
- The original TR3-style mesh swaps are used while carrying items:
MESHSWAP_MONKEY_MEDIPACK
MESHSWAP_MONKEY_KEY
- The LOT type was adjusted, matching the behavior needed for the original TR3-style pickup navigation.

SETUP:
ggg

https://github.com/user-attachments/assets/17e8a958-473a-4104-91f2-77c908b9c717


TombEngine/Game/control/box.cpp
```diff
@@ -2281,7 +2281,8 @@ void GetAITarget(CreatureInfo* creature)
 		{
 			FindAITargetObject(creature, ID_AI_AMBUSH);
 		}
-		else if (Vector3i::Distance(enemy->Pose.Position, item->Pose.Position) < REACHED_GOAL_RADIUS)
+		else if (item->ObjectNumber != ID_MONKEY &&
```
Lwmte: What this condition is doing? To me, it looks like a hack.

TombEngine/Game/control/box.cpp
```diff
@@ -2361,8 +2362,12 @@ void FindAITargetObject(CreatureInfo* creature, int objectNumber, int ocb, bool
 
 	for (auto& aiObject : g_Level.AIObjects)
 	{
+		bool matchingTriggerFlags = (aiObject.triggerFlags == ocb);
+		if (item.ObjectNumber == ID_MONKEY && objectNumber == ID_AI_AMBUSH)
```
Lwmte: Again, hacking into this function with particular object ID check is not a good way to write clean code.

TombEngine/Game/control/box.cpp
AlexxxQ: Np, I moved the condition into tr3_monkey.cpp.

The AI_AMBUSH logic was also separated between monkey and other creatures in the OG TR3 box.cpp. This condition is needed to prevent the monkey from the GetAITarget function to completion, because it change target to Lara and reset flags from AMBUSH to MODIFY - it's breaks the monkey's behavior . If this condition is excluded, when reaching AI_AMBUSH target, the monkey get the AI_MODIFY flag and, instead of dropping a SMALLMEDI_ITEM, it swaps KEY_ITEM4.

TombEngine/Game/control/box.cpp
AlexxxQ: Done, localized in the tr3_monkey.cpp.

Also TR3 OG check so that the monkey can follow AI_AMBUSH object with any bitcode.

MontyTRC89 (approved):

Lwmte (approved):
Tested both AI_MODIFY and normal mode, works correctly.
I have no objections for the code, so good to merge.

PR #2050, opened 2026-06-20
Fix JUMP_SWITCH not activating event sets

AlexxxQ:
`JUMP_SWITCH` has inverted animation states — its resting position is `SWITCH_OFF`, the opposite of normal switches. The `switchOff` check in `TestTriggers` hardcoded `ActiveState == 1`, treating a thrown jump switch as if it were resting, which suppressed its one-shot actions (event sets, cameras, flipeffects, finish, secret, soundtracks) on timed triggers.

Fixed by comparing against the switch's actual resting state (`SWITCH_OFF` for `JUMP_SWITCH`, `SWITCH_ON` otherwise), matching the inversion handling already present in `SwitchControl` and `SwitchTrigger`.

Backward compatible: normal switches are unchanged, no stored state is modified (Lua realtime reads unaffected) — only timed jump switches change behavior.


Lwmte:
Looks good to me, but can you please add a changelog entry? Thanks!

AlexxxQ:
> Looks good to me, but can you please add a changelog entry? Thanks!

Done!

Lwmte (approved):

PR #2064, opened 2026-07-17
Performance improvements for the renderer

### LINKED ISSUE #1793: Very low performance of alpha-blended and other sorted polygons
After implementation of material system, performance of sorted polygons (e.g. alpha blended ones or with any other blending mode except opaque, alpha test and additive) worsened a lot. Just having 20-50 sorted polygons in a scene can kneel down a performance from 120 FPS to 30-40 FPS.

The route of this problem lies in unoptimized way how transparent polygons are collected, sorted and rendered. Currently they are treated as a singular objects, with every new polygon requesting a bunch of state changes and texture switches.

I propose **object-based approach** for sorting and rendering polygons for objects:

- **Collection stage:** We collect sortable polygons by bucket and send a list of all polygons in that bucket to one sortable object.
- **Polygon sorting stage:** We parallelize sorting of polygons within one sortable object based on camera distance.
- **Object sorting stage:** We then sort sortable objects itself as usual, by overall center of a sortable object (this center can be precalculated bucket center).
- **Drawing stage:** We draw all group of polygons within one sortable object at once.

This way, we will do a two-stage sorting - first one will sort polygons within one sortable object, and second stage will sort sortable objects itself. 

If several sortable objects have the same overall center (i.e. two buckets of the same object with different blend modes), we may additionally sort it in this order (ascending):  exclude - subtractive - lighten - screen - alpha blend. This should correspond to expected visual results (darker areas usually appear under lighter areas in real world).

Let's imagine: I have 2 similar objects in scene, and every object has additive, opaque, subtractive and alpha blend faces. What happens is when transparent faces are collected, for every object, we create 2 sortable objects:

- Object 1 - all subtractive faces, sorted by distance from camera
- Object 1 - all alpha blend face, sorted by distance from camera
- Object 2 - all subtractive faces, sorted by distance from camera
- Object 2 - all alpha blend faces, sorted by distance from camera

After collection stage, engine sorts polygons in every sortable object in parallel, and then additionally sorts sortable objects, both by distance from camera - but in second case, it only needs to shuffle polygon groups which are already split by blend mode and sorted by distance from the camera, so sorting will be much quicker (you shuffle groups, not polygon soup).

Additionally, if sortable object buckets have the same center (let's say, object 1 subtractive and object 1 alpha blend), we draw subtractive object first, and then alpha blended - this is a fallback for cases when you can't decide which blend mode should take priority (because in real world, darker surfaces are usually below brighter ones).

This approach will limit builders ability to combine different blend modes within one object (i.e. I wont be able to make subtractive sphere within alpha-blended cube), but performance boost should outweigh this limitation
changes to existing transparency drawing code should be minimal.

Rough illustration of the approach:

Image

MontyTRC89:
## Checklist

- [X] I have added a changelog entry to the `CHANGELOG.md` file on the branch/fork (if it is an internal change, this is not needed).
- [X] This pull request meets the Coding Conventions standards: https://github.com/MontyTRC89/TombEngine/blob/master/CONTRIBUTING.md#coding-conventions

# Renderer performance: sorted faces batching, SSAO and G-Buffer order

Reduces CPU cost in the renderer, mostly in the sorted (transparent) faces pass, without changing what the pass is able to draw.

## Sorted faces

* Moved the sorted faces pass out of `RendererDraw.cpp` and `RendererSprites.cpp` into a new
`RendererSorted.cpp`.
* `SortTransparentFaces` now groups polygons that share GPU state (object instance, bucket, sprite) and assigns
each group a rank. Groups draw back to front by mean polygon distance; rooms draw first in portal traversal
order. Polygons stay depth-sorted inside their own group. Fixes #1793.
* Blend modes at equal depth now follow a fixed priority, so their relative order is stable between frames
instead of arbitrary.
* `DrawSortedFaces` accumulates the geometry of every batch and uploads it with a single map per buffer, instead
of one map per batch. When the shared buffer fills up, the accumulated batches are drawn and the remaining
geometry is retried against the emptied buffers.
* The room and object constant buffers are rebuilt only when the room, item, static or effect actually changes,
instead of once per batch.
* Sorted draws populate `Objects[0]` only, so they upload just the constant buffer prefix(`GetObjectsBufferPrefixSize`) rather than the whole ~64KB buffer. `UpdateConstantBuffer` takes an optional size for this.

## Other changes

* G-Buffer pass draws rooms front to back for early depth rejection. Other passes keep the back to front order.
* Stack reserve raised to 8 MB, which fixes a stack overflow on startup in the Lua bindings.

## Notes for reviewers

Two ordering changes are worth checking in real levels:

* Transparent room faces (water surfaces, glass) now draw before every transparent object. An object or sprite behind such a face is drawn after it.
* Sprites sharing a texture within a four block depth band batch together and draw at the mean distance of the group, so sprites of different textures mixed in the same space no longer interleave per polygon.


TombEngine/Renderer/ConstantBuffers/ObjectsBuffer.h
```diff
@@ -26,4 +27,13 @@ namespace TEN::Renderer::ConstantBuffers
...
 		ObjectData Objects[INSTANCED_STATIC_MESH_BUCKET_SIZE];
 	};
+
+	// Byte count of the CB prefix covering bones, metadata and the first `objectCount` entries
+	// of Objects[]. Draws which populate only a few instances (e.g. sorted faces always use
+	// Objects[0] with instance count 1) pass this as partial upload size to avoid re-uploading
+	// the whole ~64KB buffer on every batch. Relies on Objects[] being the last field.
+	constexpr int GetObjectsBufferPrefixSize(int objectCount)
+	{
+		return int(offsetof(CObjectsBuffer, Objects) + sizeof(ObjectData) * objectCount);
```
Lwmte: Please reformat this expression according to coding conventions - use `(int)` cast instead of `int()` cast.

TombEngine/Renderer/Graphics/IGraphicsDevice.h
```diff
@@ -70,7 +70,12 @@ namespace TEN::Renderer::Graphics
 		virtual void UnbindTexture(ShaderStage stage, TextureRegister registerType) = 0;
 		
 		virtual std::unique_ptr<IConstantBuffer> CreateConstantBuffer(int size, std::string name) = 0;
-		virtual void UpdateConstantBuffer(IConstantBuffer* constantBuffer, void* data) = 0;
+
+		// Uploads `size` bytes of `data` into the buffer (0 = whole buffer). Partial uploads let
```
Lwmte: Please remove autogenerated comments. They are not necessary.

TombEngine/Renderer/Renderer.h
```diff
@@ -296,6 +297,18 @@ namespace TEN::Renderer
 		CullMode _lastCullMode;
 		int _lastMaterialIndex;
 
+		// Room whose data is currently in the room CB during the sorted faces pass. Lets
```
Lwmte: Don't pollute Renderer.h with autogenerated comments, it's already gigantic. If you want to keep some comments, move them to where the actual usage happens, not where the variable or method is defined.

TombEngine/Renderer/Renderer.h
```diff
@@ -328,6 +341,16 @@ namespace TEN::Renderer
 
 		fast_vector<Vertex> _sortedPolygonsVertices;
 		fast_vector<int> _sortedPolygonsIndices;
+
+		// Batch of consecutive sortable objects sharing GPU state, recorded by DrawSortedFaces
```
Lwmte: Can we maybe offload the structure to a separate header file? Renderer.h is already heavy.

TombEngine/TombEngine.vcxproj
```diff
@@ -132,6 +132,7 @@
       <AdditionalDependencies>comctl32.lib;lua53.lib;bass.lib;bassmix.lib;bass_fx.lib;D3DCompiler.lib;dbghelp.lib;dxgi.lib;dxguid.lib;d3d11.lib;version.lib;liblz4_static_d.lib;sdl3.lib;spdlogd.lib;libvlc.lib;libvlccore.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <ImageHasSafeExceptionHandlers>false</ImageHasSafeExceptionHandlers>
       <LargeAddressAware>true</LargeAddressAware>
+      <StackReserveSize>8388608</StackReserveSize>
```
Lwmte, line 135: I dont fully understand what this stack size option is for.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3184,8 +3184,14 @@ namespace TEN::Renderer
...
 			}
 
-			for (int i = (int)view.RoomsToDraw.size() - 1; i >= 0; i--)
+			// In G-Buffer pass, draw rooms front to back to maximize early depth rejection. Room collector
+			// traverses portals starting from camera's room, so collection order is already front to back.
+			// Other passes keep the reversed order (back to front).
+			int roomCount = (int)view.RoomsToDraw.size();
+			for (int j = 0; j < roomCount; j++)
 			{
+				int i = (rendererPass == RendererPass::GBuffer) ? j : (roomCount - 1 - j);
```
Lwmte: Can we clarify the variable names here? `i` should remain a counter variable in a for loop, and this one should become `index`.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3797,192 +3803,355 @@ namespace TEN::Renderer
 
 	void Renderer::SortTransparentFaces(RenderView& view)
 	{
-		std::sort(
-			view.TransparentObjectsToDraw.begin(),
-			view.TransparentObjectsToDraw.end(),
-			[](RendererSortableObject& a, RendererSortableObject& b)
+		// Group sortable polygons by drawable unit instead of sorting them individually
```
Lwmte: Autogenerated comment. Remove or shorten (remove references to the issue number at least).

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3797,192 +3803,355 @@ namespace TEN::Renderer
...
+		// back-to-front); object groups follow, sorted by their mean polygon distance.
+		// Polygons are depth-sorted only inside their own group, so each group draws as
+		// a single batch instead of one batch per depth interleave.
+
+		if (view.TransparentObjectsToDraw.empty())
+			return;
+
+		// Depth step below which two object groups count as equally distant and fall back
+		// to the blend mode priority for a stable order.
+		static constexpr float GROUP_DEPTH_STEP = 128.0f;
```
Lwmte: Having constants and constexprs within a function looks dirty. To me, better option is offloading all sorted rendering functionality to a separate new module, like `RendererSorted.cpp/h`.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3797,192 +3803,355 @@ namespace TEN::Renderer
...
+				return (Primary == key.Primary && Secondary == key.Secondary && Extra == key.Extra);
 			}
-		);
-	}
+		};
 
-	void Renderer::DrawSortedFaces(RenderView& view)
-	{
-		for (int i = 0; i < view.TransparentObjectsToDraw.size(); i++)
+		struct GroupKeyHash
```
Lwmte: As above.

TombEngine/Renderer/RendererDraw.cpp
```diff
@@ -3797,192 +3803,355 @@ namespace TEN::Renderer
...
 		{
-			auto* object = &view.TransparentObjectsToDraw[i];
-			auto lastObjectType = (i > 0 ? view.TransparentObjectsToDraw[i - 1].ObjectType : RendererObjectType::Unknown);
+			size_t operator ()(const GroupKey& key) const
+			{
+				uint64_t hash = (uint64_t)(uintptr_t)key.Primary;
```
Lwmte: Coding conventions violation. use `long long` here.

TombEngine/Renderer/Graphics/IGraphicsDevice.h
MontyTRC89: Done

TombEngine/Renderer/ConstantBuffers/ObjectsBuffer.h
MontyTRC89: Done

TombEngine/Renderer/Renderer.h
MontyTRC89: Done

TombEngine/Renderer/Renderer.h
MontyTRC89: Done

TombEngine/TombEngine.vcxproj
MontyTRC89: Sol scripting compilation started suffering of stack too much low (1 MB default). Bug present also in develop.

TombEngine/Renderer/RendererDraw.cpp
MontyTRC89: Done

TombEngine/Renderer/RendererDraw.cpp
MontyTRC89: Done

TombEngine/Renderer/RendererDraw.cpp
MontyTRC89: Done

TombEngine/Renderer/RendererDraw.cpp
MontyTRC89: Done

TombEngine/Renderer/RendererDraw.cpp
MontyTRC89: Done

nite-traveler:
Overall these changes improved average FPS, but I found one scenario where FPS was reduced. I tested this in a few different levels with graphics-intensive areas, and found the following:

- Minor FPS improvement (+ 2-4 FPS) in an area overlooking a large and complex level with very high draw distance and fog
- Moderate FPS improvement (+ 5-10 FPS) in an area with a heavy use of particle effects
- Significant improvement (+ 20-25 FPS) in a large area with a lot of transparency and distant particle effects in the scene
- Moderate FPS reduction (- 8-12 FPS) in rooms with heavy use of reflective materials and mirrors

Transparency layering looked correct in most areas, but I did find two exceptions:

- A transparent texture seen through another texture with additive blend mode was layered on top
- A transparent texture seen through a transparent object was layered on top

PR #2071, opened 2026-07-23
Ring Inventory Fixes

TrainWrack:
## Checklist

- [ ] I have added a changelog entry to the `CHANGELOG.md` file on the branch/fork (if it is an internal change, this is not needed).
- [ ] This pull request meets the Coding Conventions standards: https://github.com/MontyTRC89/TombEngine/blob/master/CONTRIBUTING.md#coding-conventions

https://github.com/TombEngine/TombEngine/issues/2079
https://github.com/TombEngine/TombEngine/issues/2081


TombEngine/Specific/Input/Bindings.cpp
```diff
@@ -269,11 +269,16 @@ namespace TEN::Input
 
 	const std::string& BindingManager::GetBoundKeyName(ActionID actionID)
 	{
-		int defaultKeyID = GetBoundKeyID(BindingProfileID::Default, actionID);
-		int userKeyID = GetBoundKeyID(BindingProfileID::Custom, actionID);
-		int boundKey = (userKeyID != KEY_UNASSIGNED) ? userKeyID : defaultKeyID;
+		auto device = GetLastInputDevice();
```
Lwmte, line 272: I can't understand why custom binding profile is now replaced by a fallback to default gamepad/keyboard binding profiles in any case. Can you explain the reason behind this decision?

TombEngine/Scripting/Internal/TEN/Objects/Lara/LaraObject.cpp
```diff
@@ -583,23 +583,33 @@ int LaraObject::GetWeaponMode(TypeOrNil<LaraWeaponType> weaponType) const
 
 	switch (weapon)
 	{
-	case::LaraWeaponType::HK:
-		if (player.Weapons[(int)LaraWeaponType::HK].WeaponMode == LaraWeaponTypeCarried::WTYPE_AMMO_1)
+	case LaraWeaponType::HK:
+	{
+		auto mode = player.Weapons[(int)LaraWeaponType::HK].WeaponMode;
+
+		// If the HK weapon mode hasn't been initialized, default to Ammo 1.
```
Lwmte: In what case we can end up in a situation where HK weapon mode was not initialized? In my opinion, it's more correct to always initialize HK weapon mode to a reasonable value to avoid unnecessary hacks like this.

TrainWrack: I can intialize it in lara initialise function.

TombEngine/Specific/Input/Bindings.cpp
TrainWrack: This was needed to allow switching the return string based on the last used device. 

The custom binding always returned gamepad key before this.

TombEngine/Scripting/Internal/TEN/Objects/Lara/LaraObject.cpp
TrainWrack: I ended up using another switch and default now return rapid. Its more cleaner.

TombEngine/Specific/Input/Bindings.cpp
Lwmte: But it means that using `GetActionBinding` script function with bindable actions (e.g. action or jump etc) will always return default key name, even if player remapped it?

PR #2106, opened 2026-08-31
Fix underwater transition into dry rooms

AlexxxQ:
**The bug with the transition from the dry room to the shallows water.**

This bug occurs in waterfall setups, where the sink push her from the water room to the dry room. As you can see in the video, Lara teleport to the floor under the waterfall instead of falling.

https://github.com/user-attachments/assets/b6f3de92-8d2c-403d-a193-b37f052bae62

The reason is that the `TestLaraWaterDepth()` function snaps Lara to the shallow water floor and starts the `LA_UNDERWATER_TO_STAND` animation, cause of late item->RoomNumber update.

The solution is just add `TestEnvironment(ENV_FLAG_WATER)` condition to the `TestLaraWaterDepth()`, so Lara can't transition from the water to shallow water without physically being in the water.

### DIFF
--- a/TombEngine/Game/Lara/lara_tests.cpp
+++ b/TombEngine/Game/Lara/lara_tests.cpp
@@ -1109,7 +1109,8 @@ void TestLaraWaterDepth(ItemInfo* item, CollisionInfo* coll)
 		item->Animation.Velocity.y = 0.0f;
 		item->Pose.Position = coll->Setup.PrevPosition;
 	}
-	else if (pointColl.GetWaterBottomHeight() <= (LARA_HEIGHT - (LARA_HEADROOM / 2)))
+	else if (TestEnvironment(ENV_FLAG_WATER, pointColl.GetRoomNumber()) &&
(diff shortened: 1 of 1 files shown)


TombEngine/Game/Lara/lara_tests.cpp
```diff
@@ -1109,7 +1109,8 @@ void TestLaraWaterDepth(ItemInfo* item, CollisionInfo* coll)
 		item->Animation.Velocity.y = 0.0f;
 		item->Pose.Position = coll->Setup.PrevPosition;
 	}
-	else if (pointColl.GetWaterBottomHeight() <= (LARA_HEIGHT - (LARA_HEADROOM / 2)))
+	else if (TestEnvironment(ENV_FLAG_WATER, pointColl.GetRoomNumber()) &&
+		pointColl.GetWaterBottomHeight() <= (LARA_HEIGHT - (LARA_HEADROOM / 2)))
```
Lwmte: Please align formatting so that `TestEnvironment...` lines up with the next line `pointColl...`.

Lwmte (commented):
Please also add a changelog entry.

Lwmte (changes requested):

PR #2108, opened 2026-09-01
Adds support for animation playback for display items

TrainWrack:
## Checklist

- [ ] I have added a changelog entry to the `CHANGELOG.md` file on the branch/fork (if it is an internal change, this is not needed).
- [ ] This pull request meets the Coding Conventions standards: https://github.com/MontyTRC89/TombEngine/blob/master/CONTRIBUTING.md#coding-conventions

(Replace this sentence with links to issues this PR concerns. If not applicable, please write "N/A").

## Pull request description
Respects Next anim, next frame, end frame fields. State ID's and Anim blending are not supported.

### DIFF
--- a/TombEngine/Game/Hud/DrawItems/DisplayItem.cpp
+++ b/TombEngine/Game/Hud/DrawItems/DisplayItem.cpp
@@ -1,6 +1,7 @@
 #include "framework.h"
 #include "Game/Hud/DrawItems/DisplayItem.h"
 
+#include "Game/Animation/Animation.h"
 #include "Math/Math.h"
 #include "Objects/game_object_ids.h"
 #include "Renderer/Renderer.h"
@@ -21,6 +22,10 @@ namespace TEN::Hud
 		_prevPosition = pos;
 		_prevOrientation = orient;
 		_prevScale = scale;
+
+		// Initialize animation state.
+		RecomputeFrameData();
+		_prevFrameData = _frameData;
 	}
 
 	unsigned int DisplayItem::GetID() const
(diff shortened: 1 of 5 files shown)


MontyTRC89:
## Review: DisplayItem animations

### Breaking change: `SetAnim` / `SetFrame` behave differently in existing scripts

Every tick, the level script (`OnLoop`) runs before `g_DrawItems.Update()`. `Update()` now moves every 
display item forward one frame. Until now, scripts had to handle frames themselves, so common patterns
break:                                                                                                 

- **Static pose:** calling `item:SetFrame(n)` once used to hold the item on that frame. Now the        
animation starts playing from there.
- **Setting the frame every loop:** calling `SetFrame(n)` every loop now shows frame `n + 1`.          
`SetFrame(item:GetEndFrame())` already shows a frame of the next animation.
- **Animating by hand:** the usual pattern is `item:SetAnim(a)` followed by                            
`item:SetFrame(item:GetFrame() + 1)` every loop. `SetAnimation()` now sets the frame back to 0, so the
item gets stuck on frame 1.                                                                            

**Suggestion:** add a way to pause playback (e.g. `SetAnimPaused` / `SetAnimSpeed`), or keep automatic 
playback off by default and turn it on with a flag. Also, `SetAnimation()` should only set the frame
back to 0 when the animation number actually changes. If the new behaviour is intended, it should be   
documented as a breaking change.
                                                                                                       
### Minor
                                                                                                       
- **Allocations:** `RecomputeFrameData()` copies a whole `FrameData`, bone orientation vector included, every tick. `GetInterpolatedFrame()` also allocates a new vector each time an item is rendered. Storing only the current and previous frame indices and interpolating in the renderer would avoid both.
- **Redundant code:** the interpolation snap in `SetFrame()` / `SetAnimation()` isn't needed, because `DrawItemsController::Prepare()` already copies the current frame data into the previous one before the script runs.
- **Pause menu:** display items drawn in the pause menu don't animate. Position and orientation already behave the same way, so this isn't a regression.
- **Objects without animations (OK):** these are handled. `GetAnimData()` returns `DUMMY_ANIM`, which has one frame, and both `Animate()` and the renderer check for an empty animation list.

Please fix true problems and ignore false problems

